PR #3, opened 2018-10-14
PR requesting dispatch forward instead of redirect

cmrockwell:
I'm suggesting that you change from 302 redirect to a requestdispatcher forward call to avoid redirecting to http when behind the elb https proxy.

### DIFF
--- a/admin-base/core/src/main/java/com/peregrine/admin/servlets/AccessServlet.java
+++ b/admin-base/core/src/main/java/com/peregrine/admin/servlets/AccessServlet.java
@@ -26,11 +26,11 @@ package com.peregrine.admin.servlets;
  */
 
 import com.peregrine.commons.servlets.AbstractBaseServlet;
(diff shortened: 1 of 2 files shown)


schaefa:
Thanks for your contribution.

I am currently working on a fix that does not require to do a redirect by doing an Intra-Sling call.

cmrockwell:
I'm not familiar with intra-sling calls. I look forward to learning about that.  With that alternative solution, I hope you'll consider the general case that PR seeks to address.
Since Peregrine CMS encapsulates and defines how it will allow servlets work in AbstractBaseServlet, I am suggesting the general ability to do regular JavaEE request dispatching. As it is, the ForwardResponse class in AbstractBaseServlet only allows dispatching to Sling Resources, but that is too narrow in my view since Sling and Java also have servlets registered on paths (such as /system/sling/info). Without this kind solution, the problem could come up again you find another scenario in where requests need to be dispatched to path registered servlets.

PR #31, opened 2019-02-27
Feature/material range

dna113p:


reusr1:
http://localhost:8080/content/admin/objects.html/path///content/objects/example (sample) still seems to have the old slider

dna113p:
Updated the dialog for the sample object, should be fixed now (styles will not look quite right but fixed in another branch)

Issue #42, opened 2019-03-18
Newly-created site shows wrong label on folders


devintuffy:
To reproduce:
- Navigate to 'sites', use + button to create a new site
- Go through the wizard, give site a name. Use "themeclean" as the basis

Result:

While the breadcrumbing shows the "correct" name of the folder or item, the explorer shows the site name for every sub-item in the site:

<img width="327" alt="Screen Shot 2019-03-18 at 1 06 11 PM" src="https://user-images.githubusercontent.com/46942507/54560089-c74e1180-497e-11e9-9a5e-d5e8d2f57425.png">

<img width="496" alt="Screen Shot 2019-03-18 at 1 06 20 PM" src="https://user-images.githubusercontent.com/46942507/54560232-18f69c00-497f-11e9-9826-91d4f07ffa5d.png">

devintuffy:
Oh nooooo

![Screen Shot 2019-03-18 at 1 10 31 PM](https://user-images.githubusercontent.com/46942507/54560284-3f1c3c00-497f-11e9-9c05-247eea348209.png)

devintuffy:
Looking at the response from http://localhost:8080/perapi/admin/nodes.json/content/sites/test%20area/Sample%20Sites/site%202/child%201, I can see that this 'title' value is coming from the server:

![Screen Shot 2019-03-18 at 2 04 16 PM](https://user-images.githubusercontent.com/46942507/54563615-e3ee4780-4986-11e9-8ada-c283cab58e48.png)

devintuffy:
Composum confirms this is persisted:

<img width="962" alt="Screen Shot 2019-03-18 at 2 19 03 PM" src="https://user-images.githubusercontent.com/46942507/54564352-d0dc7700-4988-11e9-8df1-1043e1f8b912.png">

devintuffy:
This is happening in the createSite api call:

When creating a new site from a template, we call com.peregrine.admin.resource.AdminResourceHandlerService#copySite, with the new site name as "targetName". This targetName is passed into com.peregrine.admin.resource.AdminResourceHandlerService#copyChildResources, which copies each page and then updates the title -- so in this case, updating every page's title to the same thing.

Luckily, copySite is the only thing that calls copyChildResources, so fixing shouldn't be difficult.

PR #43, opened 2019-03-19
Fix item titles when created site from template

devintuffy:
Fix for #42. When calling copyResources to copy a site from a template, only use the new name for the top-level resource, to prevent every object from having the same name.


reusr1 (changes requested):
there seem to be some problems here. I started a new site with the tile 'ruben' and found some issues in the resulting structure: http://localhost:8080/bin/browser.html/content/sites/ruben/Sample%20Sites/jcr%3Acontent

the resulting properties contain null as part of the path structure resulting in pages not rendering (eg for example compare the themeclean templates with the newly created site templates)

reusr1 (approved):

Issue #41, opened 2019-03-18
Site, page other Wizards: user sees validation error immediately upon page load


devintuffy:
- Sites -> New Site

<img width="1417" alt="Screen Shot 2019-03-18 at 12 58 56 PM" src="https://user-images.githubusercontent.com/46942507/54559596-9f11e300-497d-11e9-9de2-89081956ba7c.png">

"selection required" error is displayed right away. This should only appear if the user attempts to click "next" without selecting.

UI tweaks to fix with this as well: 
- Fix capitalization on "selection required"
- Add some sort of treatment so it is obvious to user that the theme area is clickable.

PR #44, opened 2019-03-19
Issues/41

devintuffy:
Fix for issue #41. 

Existing code bound the error display directly to the form model, resulting in validation being triggered right when the page loads and displaying a big red “PLEASE SELECT A VALUE” error message to the user.

This fix moves that validation to separate validation objects which only get triggered when the user moves forward.

### DIFF
--- a/admin-base/materialize/custom/_vue-form-generator.scss
+++ b/admin-base/materialize/custom/_vue-form-generator.scss
@@ -34,7 +34,6 @@
     box-sizing: border-box;
 
     .collection {
-        border: 0;
     }
 
     .field-wrap {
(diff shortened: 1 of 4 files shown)


reusr1:
there seems to be an issue with single selects - they are not selected (but in multi select the first one is selected) and the style for single select lists seems to be off: 

image

reusr1 (changes requested):
please see previous comment

devintuffy:
Style problem is cause by a conflict between vue-form-generator and materializecss:

Screen Shot 2019-03-20 at 8 35 34 AM

We added this "border: 0" in custom/_vue-form-generator.scss, will adjust so it doesn't break other styles

reusr1 (approved):

PR #45, opened 2019-03-20
Bugfix/multiselect

dna113p:
Fixed several issues with editing content with subchildren (collections and multiselects). 

- The server was looking up items using a "name" property which breaks when component also uses a "name" property. 
Fixed by changing resource lookup to use "path".

- Items are deleted by sending to the server with an "_opDelete" property set on them. Previously this was managed using the same model that vue-form-generator uses to render the page, causing multiple rendering bugs. 
The fix was to create a separate "_deleted" object in our "/state/tools" object. Whenever editing a component, deleted items should be moved to this object and then merged back in only when the data is ready to be sent to the server. This allows vfg to cleanly maintain the model.


reusr1 (changes requested):
there is an issue with reordering in the dialog

reusr1 (approved):

Issue #51, opened 2019-04-01
Explorer UX updates


devintuffy:
The UX for the explorer is not as intuitive as it could be. Two particular problems:
- It is not obvious to the user that the rows can be reordered
- Users expect Edit functionality when clicking on a label, not drill-down.

To fix this:
- Add drag handles to the left of the rows. For now we will use the materializecss drag_handle icon. The cursor should change to "grab" and "grabbing" while rows are being moved.
- Change the functionality for single-click on an asset name to display the edit window; add a double-click handler to drill down. We should also add a second way to drill-down that doesn't require a double click, either  drill-down icon to the right of the label or a folder icon to the left.

devintuffy:
Pull request: #52

PR #52, opened 2019-04-02
Issues/51

devintuffy:
PR for ticket #51 

In this branch:

- Style updates for explorer: added drag handle, changed cursor to indicate dragging (please note that Chrome and Firefox don't support css changes to cursor during drag, but Safari works nicely), some other minor CSS tweaks

- Change in behavior of explorer navigation: clicking on an item name edits the item (if applicable), added folder icon to drill down into folder.

- (currently turned off) Folder sub-item count badge display to show users how many items are in a folder. We need to figure out the best way to display this number, right now I am showing them as overlay numbers on the folder, but that UX is more suited to 'new item' that can be cleared. To see this work, change the call to writeProperties on line 130 of NodesServlet.java from writeProperties(child, json, false) to writeProperties(child, json, true);

If we decide not to use the display count, I'll remove the code that counts children.

### DIFF
--- a/admin-base/core/src/main/java/com/peregrine/admin/servlets/NodesServlet.java
+++ b/admin-base/core/src/main/java/com/peregrine/admin/servlets/NodesServlet.java
@@ -26,6 +26,7 @@ package com.peregrine.admin.servlets;
  */
 
 import com.peregrine.commons.servlets.AbstractBaseServlet;
+import com.peregrine.commons.util.PerConstants;
 import com.peregrine.commons.util.PerUtil;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ResourceResolver;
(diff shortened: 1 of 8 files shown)


reusr1:
I do like how it feels. I am not sure I like the count overlay - maybe a difference in the icon itself may work (eg show empty vs populated folder). I do also think if we need the child count we may be better off adding this functionality to the PerBase class and adapt the resource to it.

devintuffy:
Further updates
- Added and implemented dblClickCommand to admin-components-action. If this is specified, we will manually count the timing of successive clicks and if two or more are fired within the timeframe we call the doubleClick action, otherwise the default. If no dblClickCommand is specified then we don't bother with any of that and just call model.command as usual.

- removed the child count code until we figure out the best way to fetch/display it

devintuffy:
The doubleclick functionality leaves a slight delay before the editPage starts. That would be unavoidable, after a single click we need to wait a period of time (currently set to 200ms) to see if another click happens before proceeding with the action. In practice it feels like more than 200ms. I'll play with this a bit more, I don't really care for the experience as-is. Might just be a matter of timing adjustments.

Update after a little more experimenting: the additional delay I am noticing might be the editor opening, as the direct edit button (which doesn't have the doubleclick attached, so no delay) also has a slight pause. 

We may want to add in a "loading" state for actions with a delay. Like a slightly opaque overlay with some simple progress animation? That would be for another ticket.

PR #53, opened 2019-04-03
Bumped up version of directory-maven-plugin from 0.1 to 0.3.1

00v3rdr1v3:
Bumped up version of directory-maven-plugin from 0.1 to 0.3.1 in parent/pom.xml


reusr1:
thank you for the PR

Issue #54, opened 2019-04-04
special characters in node creation


reusr1:
we need to make sure when we create nodes (new page, new template, new object, new asset) that we handle special characters correctly. As it stands right now creating a page with a question mark in the name results in a node that can break peregrine. 

In general we should keep the display name as typed and change non alphanumeric characters to a dash and make sure we do not have a conflict when creating such a node (does a node exist already with the node name we are trying to create)

devintuffy:
Also with this ticket I will fix handling of duplicate names. If the user re-uses  name, they get a notification that the name is already in use:

<img width="1077" alt="Screen Shot 2019-04-08 at 1 10 05 PM" src="https://user-images.githubusercontent.com/46942507/55753630-ab7ae000-59ff-11e9-94ae-40c62804df0c.png">

which is a great first step, but it doesn't block the input; the user can click through to the end of the wizard where they generate a javax.jcr.ItemExistsException and the page breaks.

As part of this ticket, will block user from proceeding forward if a name is in use. 

This will also take into account the escaping done as described above; so if a user previously created an item named "hello?123" that gets persisted as "hello-123", anything that would eventually persist the same way should also trigger the same name-already-in-use error, e.g., "hello?123", "hello%123", "hello-123"

devintuffy:
Pull request for this: https://github.com/headwirecom/peregrine-cms/pull/61

PR #61, opened 2019-04-16
Issues/54

devintuffy:
Handles issue #54 

In this PR:

- Node names are filtered to only allow URL-friendly characters, but we now distinguish between the node name and title/"display name". Node names can only have alphanumeric and a few other URL-safe characters, while display name can contain any unicode characters. When a user types in a name containing special characters, these are filtered out for the URL values, but in the explorer and elsewhere they will see the name as they typed it. e.g., if a user names a node "hello, world?", the node will be internally named "hello--world-", but the user will only see "hello, world?" outside the URL and composum.

- Combining characters are gracefully handled when filtering the node name; instead of dashes, the closes ISO-8559-1 character is used where possible; e.g., "hēllo?" becomes "hello-"

- The "name available" box on the page wizard now blocks moving forward if the name is taken. This check is aware of the updates above, so if there is a node named "hēllo?" and the user enters "hêllo&", this will block, as "hēllo?" and "hêllo&" would both become "hello-" when the node is written.

A few things to note here:

1. Per https://sling.apache.org/documentation/the-sling-engine/request-parameters.html#character-encoding, sling expects forms to have a "\_charset\_" field; the sling servlet will look for this and interpret parameters based on the encoding. Since apiImpl mainly calls servlets by manually creating a FormData object and posting it, I created the util class UTF8FormData in utils.js, which is just a FormData object with \_charset\_ set to UTF8. 

2. To handle accented and other combining characters in a way that kept the URLs readable, I normalized the text to split it into base characters and diacritics, then filtered out the latter. See https://docs.oracle.com/javase/tutorial/i18n/text/normalizerapi.html for info on the first step of that. I  implemented this both in Java and in Javascript -- I think that the job of handling such characters should be done primarily on the server side; the client should only send one name parameter (as opposed to the client sending separate 'nodeName' and 'displayName' params) and let the server sanitize as needed. However, when we are using javascript to check if a name is a duplicate, we need to know how the server will be massaging the data, so the same method is needed on the front end.

### DIFF
--- a/admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandler.java
+++ b/admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandler.java
@@ -1,11 +1,9 @@
 package com.peregrine.admin.resource;
 
-import com.peregrine.transform.ImageContext;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ResourceResolver;
 
 import javax.jcr.Node;
-import javax.jcr.RepositoryException;
 import java.io.InputStream;
 import java.util.Map;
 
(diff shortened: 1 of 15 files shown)


reusr1:
we need to prevent someone from creating a page or object that contains a dot in the name as well - if you create a page called a.a and then try to view it that will not work due to sling and the concept of selectors

devintuffy:
> we need to prevent someone from creating a page or object that contains a dot in the name as well - if you create a page called a.a and then try to view it that will not work due to sling and the concept of selectors


Oooh, did not know about the . character. I'll revise that regex, right now it allows these in addition to alphanumeric: $ \ - _ . + ! 

To be safe, I'll restrict that to alphanumeric and - and _

PR #76, opened 2019-05-22
feature/additional_viewports

flx-sta:
Add 3 additional Viewports:
- mobile-landscape (576px)
- tablet-landscape (1024px)
- laptop (1366px)

Screenshot 2019-05-21 at 15 58 00


reusr1:
thank you for the contribution. it seems the change makes the toolbar look very busy. I wonder if a different UX approach such as 'rotate' would be better or a way to define a list of viewports for a site instead of having a hard coded list. @devintuffy also mentioned the other day that we should find a better way to separate the preview icon from the viewport size icons.

flx-sta:
Yeah we had the same thought. 
We are currently working on a different approach where you have a separator between preview and viewports and where the viewports are grouped as a drop-down menu.
If you have a better idea feel free to share

flx-sta:
There is something missing in this PR.

The buttons are only available on the page-editor but not on the template-editor

flx-sta:
Buttons are now available in the template-editor too

reusr1:
@bkheadwire can you please review this PR and merge if it all checks out? Please also run the e2e tests to see if they need adjustments based on this PR.

bkheadwire:
@reusr1 the new views work. The e2e test currently doesn't touch the viewports so it's unaffected by this change. One thing I'm unclear on from reading the comments here is whether we want to wait for the viewport buttons to be organized more cleanly before merging this or if we want to tackle that as a separate issue later.

flx-sta:
@bkheadwire there is already a pull request in the pipeline for that: #120
I’ll tell m colleague to remove it from „draft“ and make it an actual request

PR #90, opened 2019-06-14
Mark template components in page-editor

flx-sta:
This feature adds a background-image to template components in the page-editor so its easy to directly determine them instead of having to click on them to get the notice.

Screenshot 2019-06-14 at 09 23 05


dna113p:
Awesome this is really nice to have! :)

Issue #94, opened 2019-06-19
i18n is case sensitive but folders arent


flx-sta:
While I was working on the german translation for peregrine I had to realise, that the key (which is the english word) for translation is case sensitive but creating folders in the [`i18n.admin.de`](https://github.com/headwirecom/peregrine-cms/tree/develop/admin-base/ui.apps/src/main/content/jcr_root/i18n/admin/de) directory is not.

For example you have the following 2 code lines
```
<span>{{ $i18n('Create') }}</span>
<span>{{ $i18n('create') }}</span>
```
You could either translate Create or create. But not both since you can't create a directory "Create" and then "create".

bkheadwire:
@reusr1 this was already semi-fixed as part of [PR #95](https://github.com/headwirecom/peregrine-cms/pull/95) - all keys are lowercased before requesting translations. However, this effectively means that anything that goes through i18n ends up coming back with whatever capitalization is in the i18n tree regardless of what it looked like initially. Thoughts?

flx-sta:
The current fix is not a really a fix more like a workaround.

I would rethink the concept and think about using actual unique keys since this is the more common used concept

e.g: 
```
// de.properties
EXPLORER_BUTTON_REMOVE=entfernen

// en.properties
EXPLORER_BUTTON_REMOVE=delete
```

flx-sta:
So I came to understand thats it is actually possible to use a key-approach (based on file-names though).
Just put the key into the `i18n/de` and `i18n/en` directories

flx-sta:
See #297

PR #95, opened 2019-06-19
german translation (& i18n non case sensitive)

flx-sta:
This adds a couple of translations for the german language.
They might not always be perfect.

A lot of code had to be change to make a translation possible.
Especially the `it18n.js` file where i made the key parameter case insensitive.
See Issue #94 for more information about this.


reusr1:
@bkheadwire please review and make sure the e2e tests still work

PR #104, opened 2019-06-25
Extending Page with additional properties 

NMuzaffar:
The primary task is to extend base per:Page with additional properties.

- A mixin **per:PageProperties** is created and multiple properties are added (Sitemap/Robots.txt related properties might be added in the future to manage sitemap/robots through admin console. That is why it is decided to create new mixin.)
- Nodetype **per:PageContent** extends from **per:PageProperties**
- Mixin **per:PageProperties** contains the following properties: 
         1.  protocol (_string_) -> default value **"http://"**
         2. hostname (_string_) -> default value **"www.example.com"**
         3. canonicalLink (_string_) -> default value **Canonical Link of current Resource page** from 
                   custom **UrlExternalizer** service
         4. excludeFromNavigation (_boolean_) -> default value **FALSE**
- **UrlExternalizer** service to externalize URLs when generating canonicalLinks for pages. When building externalized url, the sling run mode is checked for author, publisher runmodes through SlingSettings service. If none found, defaults to local instance. 
- **PageEventHandlerService** which handles the aforementioned properties' values (E.g. if the page resource is moved to a different place and it has children pages, the page properties are checked/updated. For example canonicalLinks are updated across all change Pages.)


reusr1:
hi, it seems this PR is mixing a couple of issues together. Is a per:PageProperties  node type really the right answer here? 

For some aspects of this PR it may be better to define a per:SiteRoot node (for example if /content/sites/{tenant} is a per:SiteRoot either as a mixin or as a full node (not allowing editing as a page anymore). 

I am also wondering a bit what the canconicalLink is for - do you have some use cases?

PR #116, opened 2019-07-05
pathbrowser: browsertype - image

flx-sta:
### Description
I've added a new browserType to the pathbrowser called `image` where you can only select images.
Its default value for the `withLink` option is `false` because images should normally be available locally instead of referencing external images.

### Images
Screenshot 2019-07-05 at 10 00 41 Screenshot 2019-07-05 at 10 00 48 Screenshot 2019-07-05 at 10 00 57

### Additionals
I've added a new constant for admin called `PathBrowser`:
```
export const PathBrowser = {
  Type: {
    PAGE: 'page',
    ASSET: 'asset',
    IMAGE: 'image'
  }
}
```
_I wasn't happy with the fact using the same string over and over again. That's inconsistent._


reusr1:
@Felix-Puetz sorry for the late review. there is a common use case to have assets somewhere else if the site owner uses a different DAM (for example cloudinary or the adobe aem assets solution) We have to make the use of links to outside assets optional.

flx-sta:
@reusr1 I'm not quite sure if i know what you are referring to.
I haven't removed the "link" function. I've just made it, as you say, optional.
You can still pass the `linkTab` parameter with a value of `true` to have it enabled

reusr1:
looks good from my side - we will need to adjust the fragments in themeclean accordingly as well as add this field to the percli hatch command and the related documentation before fully merging.

dna113p:
Do we want to add more to this? For example you may want to find just PDFs, or just SVGs. In themeclean for example we have a pathbrowser that should only allow video.

flx-sta:
@dna113p Agree though we could merge this and create an issue (feature tag) to remind us that this should be extended to what you mentioned

PR #120, opened 2019-07-08
Encapsulate screen width icons into a dropdown menu

felixulbrich:
- Added a materialicondropdown component that resembles a dropdown menu, represented by the materialicon of the selected item (default value possible)
- Encapsualted the screen width icons into the new dropdown component
- Adjusted z-index for workspace so the dropdown is visible in preview mode (nav and workspace shared the same z-index before the change)
screen_width_dropdown

- Might consider https://github.com/headwirecom/peregrine-cms/pull/76 then


flx-sta:
@felixulbrich please take this pull request out of Draft Mode and make it read for review

reusr1:
@felixulbrich can you please make sure the .travis.yml file is not included in your PR?

felixulbrich:
@reusr1 Done.

reusr1:
sorry for the long time it took us to look at this PR. I think it's a good idea, however I feel we already have a dropdown component (material-select) and we should really be reusing that one also from a UX perspective. The current UX looks a bit odd to me.

flx-sta:
Does the material-select provide icon-only drop-down?

reusr1:
@Felix-Puetz , I dont think we support that right now but I think it would be better to change that as it is a more general approach than writing a one-off component

flx-sta:
@reusr1 Alright I understand. I'll see what we will do with that, since the current approach does fullfill our current needs so the clean approach might take longer to be ready

flx-sta:
@reusr1 it's a litte bit confusing though. material-select is defined as a "field" and technically the viewport drop-down is not a field, because its not a configuration for a component. 
It's part of the subnav.
What about that?

flx-sta:
So, I tried to do it with a vue-multiselect and it's just a mess.
You literally take the whole multiselect styles and rip them apart to get what we have right now. 
Additionally its not really possible to put an icon-action into the drop-down and make it work. 

Thats due to `<action/>` having the `@click.stop.prevent` modifiers



I've stoped working on the approach with multiselect and ask again: Does it actually make sense to do it that way instead of just keeping it as it is?
Screen Recording 2019-08-14 at 16 26 27

flx-sta:
Push
@reusr1 @dna113p @GastonGonzalez @liqin @polished85

bkheadwire:
Hi @Felix-Puetz,
I checked out this branch, merged develop into it, and tested it on a brand new peregrine instance. A couple of things:

1. I see the view options both in their original place (to the left of the preview button) and in the dropdown. Is this intended?
01-buttonsanddropdown

2. There's a console error upon loading the page editor:
02-consoleerror

reusr1:
@Felix-Puetz https://github.com/headwirecom/peregrine-cms/tree/feature/actionmultiselect contains an alternate implementation of the drop down. Just an experiment, would love to hear what you think.

flx-sta:
ezgif com-video-to-gif (5)

@reusr1 I kinda like your approach.. especially the small touch with the bottom part of the circle changing to a rectangle.
Its missing the other, new screen sizes though

reusr1:
@Felix-Puetz great to hear that you like it - I'll merge it as is and then we can add the other screen sizes back?

flx-sta:
I'd appreciate if we add them right away so we can merge it into our fork asap too

flx-sta:
### Closing Reason
The feature has been implemented and refined in branch [feature/actionmultiselect](https://github.com/headwirecom/peregrine-cms/tree/feature/actionmultiselect)

PR #135, opened 2019-07-11
issues/133

devintuffy:
pass target component data to beforeStateActions; use this to fix the 'save edits?' functionality after clicking cancel.


reusr1:
this PR is not necessary anymore as the issue is addressed in the feature/actionrewrite branch

PR #147, opened 2019-07-15
Issue/124

GastonGonzalez:
Hi Ruben -

This is a bare bones implementation that works with k8s/docker. These other repos were also involved in this implementation:

- https://github.com/peregrine-cms/peregrine-k8s-explore (already in master)
- https://github.com/peregrine-cms/peregrine-docker-apache-explore (already in master)

Items remaining
- Add deactivation support
- Set activation status
- Invalidate references
- Create k8s runmode so that we can include sample OSGi configuration

### DIFF
--- a/platform/replication/core/src/main/java/com/peregrine/admin/replication/impl/ModPageSpeedCacheInvalidationService.java
+++ b/platform/replication/core/src/main/java/com/peregrine/admin/replication/impl/ModPageSpeedCacheInvalidationService.java
@@ -51,8 +51,6 @@ import java.util.List;
 
 /**
  * This class is responsible for issuing cache flush requests to mod_pagespeed on replication events.
- *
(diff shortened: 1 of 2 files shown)


platform/replication/core/src/main/java/com/peregrine/admin/replication/impl/ModPageSpeedCacheInvalidationService.java
```diff
@@ -0,0 +1,211 @@
...
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * This class is responsible for issuing cache flush requests to mod_pagespeed on replication events.
+ *
+ * @author <a href="mailto:gg@headwire.com">Gaston Gonzalez</a>
```
reusr1: probably better to not leave @author in the code (due to possible spam)

PR #163, opened 2019-07-29
changed width of placeholder to auto

FabianHaupt:
We found that the placeholder was going out of the screen on the right side.
By changing its width from 100% to auto it fixed the problem and didn't cause any new ones.


reusr1:
@bkheadwire can you please check this PR out? thank you

PR #200, opened 2019-08-22
 Added a Title component in themeclean, referred from AEM core component library 

tsaurabh74:


tsaurabh74:
The reference of the component is taken from the Adobe Core Component library

https://opensource.adobe.com/aem-core-wcm-components/library/title.html

So IMO, we should have this component in our library too.

Thanks!

reusr1:
thank you for your contribution - have you had a look at https://github.com/headwirecom/themeclean-flex/tree/develop/fragments/simpletext ? For performance reasons we are moving over to that theme in the near future and not expanding themeclean anymore (as it uses bootstrap, jquery with a heavy initial load penalty). In general your component looks ok, however I question if it would be better to allow for the H element to be configured and to remove the outer div as that adds additional unnecessary html elements

tsaurabh74:
Thanks for your suggestion, I am closing this PR and created a new PR.
Below is the link:-
https://github.com/headwirecom/themeclean-flex/pull/73

PR #201, opened 2019-08-22
fix a bug where templates cant get access tags from multiple objects

FloMarks:
Templates can only access tags from a single source e.g.:
"valuesFrom": "/content/objects/blog/authors.infinity.json"

In this fix it is possible to pass multiple paths, so the template can accept tags from multiple sources e.g.:
"valuesFrom": [
"/content/objects/tags.infinity.json",
"/content/objects/tagstwo.infinity.json",
"/content/objects/tagsthree.infinity.json"
]


FloMarks:
Pull request closed because the functionality is no longer needed.

Issue #194, opened 2019-08-20
admin component templateview and pageview should expose same UX concepts


reusr1:
pageview and templateview (editing of page level details and template level details are not consistent at the moment. They should both use the pageview UX (icons, ok to save)

flx-sta:
I was just about to write the same issue 👍

PR #198, opened 2019-08-22
Meta-tags and tabs for site-info and tags

NiklasFLuis:
### Description

Add tabs for site-information and meta-tags aswell as meta-tags.

### Images
site-info-edit
site-info
meta-tags-edit
meta-tags

PR #202, opened 2019-08-23
unify page- and templateview

flx-sta:
Add a new component `<explorerpreviewcontent>` which is an abstract representation of pageview and templateview.

fix for issue #194


reusr1 (dismissed):
@Felix-Puetz it seems your change alters the style a bit and is not in line with the objects and assets preview (borders around the actions missing due to changing from li to a tag). I am also not sure about the icon on the left (it may be a good idea, however, then the assets and objects view need the same UX for consistency)

flx-sta:
> @Felix-Puetz it seems your change alters the style a bit and is not in line with the objects and assets preview (borders around the actions missing due to changing from li to a tag). I am also not sure about the icon on the left (it may be a good idea, however, then the assets and objects view need the same UX for consistency)

@reusr1 I had this in mind and thought about taking object-view and asset-view unification (as present here) in as issue for the future.
About the missing lines and the icons I'd appreciate screenshots to be able to get more into detail

reusr1:
> > @Felix-Puetz it seems your change alters the style a bit and is not in line with the objects and assets preview (borders around the actions missing due to changing from li to a tag). I am also not sure about the icon on the left (it may be a good idea, however, then the assets and objects view need the same UX for consistency)
> 
> @reusr1 I had this in mind and thought about taking object-view and asset-view unification (as present here) in as issue for the future.
> About the missing lines and the icons I'd appreciate screenshots to be able to get more into detail

Felix,
 
image has no lines between the icons vs what you see in assets: 
image

flx-sta:
Oh I see.

And i guess with "the item on the left" you are refering to the "gear" icon.

Is there a problem to adapt this layout? Adding the lines should not be too big of a deal

reusr1:
> Oh I see.
> 
> And i guess with "the item on the left" you are refering to the "gear" icon.

yes. the lines are not rendering because the items are not <LI> items anymore

> 
> Is there a problem to adapt this layout? Adding the lines should not be too big of a deal

it's ok - the icon seems a bit massive and why is it not in the corner (I am not a designer :-)?

flx-sta:
I've created a task for these changes in our System 👍

reusr1 (dismissed):
it seems this PR does not work correctly when merged with develop - all page/template/assets/object properties are missing. May be realted to the metadata tab as that is not in develop

flx-sta (commented):
👍

flx-sta:
@reusr1 finish merging pr/meta-tags #198 into this one

pagerenderer/vue/ui.apps/src/main/content/jcr_root/apps/pagerender/vue/structure/page/og_tag_dialog.json
```diff
@@ -0,0 +1,68 @@
...
+          "value": "website"
+        }
+      ]
+    },
+    {
+      "type": "pathbrowser",
+      "browserRoot": "/content/assets",
+      "browserType": "image",
+      "label": "og:image",
+      "readonly": true,
```
reusr1: @Felix-Puetz any reason why the og:image property would be read only? seems to not make any sense

reusr1 (changes requested):

pagerenderer/vue/ui.apps/src/main/content/jcr_root/apps/pagerender/vue/structure/page/og_tag_dialog.json
flx-sta: We used to do that to force the user into using the pathbrowser. But in the end we had to undo this because its nonsense. So we can remove that

reusr1 (approved):

reusr1 (dismissed):
looks like the image preview is now splitting the image and the data into a 50/50 split - does not seem that nice as we loose a lot of screen real estate this way

flx-sta (commented):
👍

flx-sta:
~~I don't quite understand why its having conflicts. It's telling me "up to date" when i try to merge develop~~

Nevermind. It was not properly fetched

flx-sta:
Please make sure these changes will not break anything due to some og-tag stuff being merged by accident

PR #203, opened 2019-08-23
Add material-select for the defined site-language

NiklasFLuis:
### description
Add a material-select to the site-information to set the `lang`-attribute of current site. Available choices are 'en' and 'de'. Only the choice from the template is used.

### images
lang-select
lang-choices

lang-tag


reusr1:
while looking at this a bit more I think it may be better to define the site language only at one level and use inheritance to drive the current site language instead of setting it per page (dropdown would have default and the the languages as a selection). To prep for the future it would also be better to populate the available languages from a servlet instead of a hard coded list. Let me know what you think.

flx-sta:
@reusr1 as discussed: please provide a inheritance solution to replace this one.

and inform us when its implemented since we want to merge it into our fork too 👍

reusr1:
created issue #271 to define a solution that fits all uses cases for the language definition in peregrine

Issue #196, opened 2019-08-21
Including of external javascript file for Internet Explorer


andreas-hauser:
The shim.min.js file is included from an external source (it seems to be necessary for the Peregrine editor to work under Internet Explorer):
```javascript
if(/MSIE \d|Trident.*rv:/.test(navigator.userAgent))
   document.write('<script src="https://raw.githack.com/zloirock/core-js/v2.4.1/client/shim.min.js"><\/script>');
```
We might want to avoid including external sources, as these could come with cookies, etc.
Would it be a good idea to provide the specific javascript file within peregrine?

flx-sta:
@reusr1 @GastonGonzalez @dna113p 

What's your opinion on that?

reusr1:
We can insource it for sure. Should follow the same process as all other insourcings

reusr1:
file is now insourced

PR #204, opened 2019-08-26
Insource external shim.min.js file

andreas-hauser:
Insource external shim.min.js file (necessary for the Peregrine editor to work in IE) as described in issue #196. Usage of the insourced shim.min.js file for the available sample pages is also added.


reusr1:
hey, thank you for this PR - unfortunately we do have to change the approach on this a bit. The page renderer should not rely on the admin console as the admin console may not be necessary on some instance types (say a server only serving a site and not the admin console). It would also be good to pull the shim from the npm repo instead of just copying it into the project.

andreas-hauser:
@reusr1 
as the file is only needed by IE, is there a good way of importing it from the npm repo only when necessary?

reusr1:
@andreas-hauser - have a look at the maven pom file @ https://github.com/headwirecom/peregrine-cms/blob/develop/pagerenderer/vue/ui.apps/pom.xml - this instructs maven to copy the files to the final package that is installed in peregrine. The shim should be added to that list but needs to go into a different folder in order to not be included in the combined js file. The reference to the file from the page components can then stay the same, just insourced.

andreas-hauser:
@reusr1 
the shim has been moved into the pagerenderer module and it's included in the new conditional-js folder.

PR #210, opened 2019-08-28
german translation for templates

FloMarks:
deposit a german translation for multiple buttons when working with templates. For more details see commit messages.


admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/subnav/template.vue
```diff
@@ -27,6 +27,19 @@
...
+                v-bind:model="{
+                  command: 'selectPath',
+                  download: getDownloadPath(),
+                  target: getPath() + '/jcr:content.xml',
+                  tooltipTitle: $i18n('Export Module'),
+                  title: 'Export',
+                  type: 'download'
+                }"
+            ></admin-components-action>
+        </template>
```
reusr1: question on this addition: where would this show up and what does it do? also it seems the isEditor() method is not defined or am I missing something here?

reusr1 (dismissed):
this PR assumes that the dropdown for the page view is already merged - this can not be merged into develop as is. there are also some off files preventing an installation of the admin-base/ui.apps package to be successful

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/subnav/template.vue
flx-sta: You're right. Somehow it got lost on the way... because in my repository the method is there..I'll have to fix that

Felix-Puetz:
@reusr1 as discussed: please provide a inheritance solution to replace this one.

and inform us when its implemented since we want to merge it into our fork too 👍

flx-sta:
Alright this should be fine now.... did some tweaks

PR #211, opened 2019-08-28
PageMerge: sonar fixes + coverage

reggie7:


reggie7:
Let me recheck it on Monday to confirm it does not break anything. I'll confirm here that this is indeed the case.

schaefa (changes requested):
- The same issue with formatting as in 213
- Please make sure that you run the IT tests (go to /testing and run it with 'mvn clean instal l-P standalone') are passing. 

The rest looks fine to me

reggie7:
Stuff fixed, I can't confirm `mvn clean install -P standalone` though - it does not work for me in `develop`. Please check if it still works on your side.

reggie7:

PR #213, opened 2019-08-29
Sonar fixes: com.peregrine.pagerender.vue.models.PageModel

reggie7:


schaefa (changes requested):
Here are a few things:
- do not replace imports with .* (it reduces the readability)
- do not change the order of methods or variables (makes reviewing code difficult even later when checking out Git History) or their formatting
- Add a comment to Optionals (I like them but they are hard to read until we all get comfortable)to what they mean like 'get grant parent of resource and if Resource Type is PAGE_PRIMARY_TYPE return its JCR Content node if found)

The rest is fine as far as I can see

reggie7:
This one is a small PR fortunately, so all the more I'd leave the reordering of methods made. There was a method and a constructor in between the fields, it's unaesthetic :(

Regarding `Optional` - I'm thinking it makes [little sense](https://twitter.com/unclebobmartin/status/870311898545258497) to comment those, one would easily get used to the notation, I know from experience (I liked the old school for readability, but got used to `stream` notation quite quickly after all).

reggie7:

PR #215, opened 2019-09-04
Sonar fixes: com.peregrine.commons.util.PerUtil

reggie7:
Plus coverage of the code changed in the process.

Most of the changes shown are from renaming a constant: `EQUALS` => `EQUAL` and replacing `PerUtil.isEmpty` and `PerUtil.isNotEmpty` with its `StringUtils` equivalents (`import` gets replaced). After that only a handful of files remain to check.


reggie7:

PR #221, opened 2019-09-09
com.peregrine.commons.util.PerUtil

reggie7:
Replacement for [215](https://github.com/headwirecom/peregrine-cms/pull/215).


schaefa:
I merged all your changes from this PR into develop-sling11.
Due to my decision to have the common Mock code in commons-test I had to move any test from commons into commons-test. This is not optimal but I have no other solution.
We should also find a solution to replace Guava in commons as I only used two methods and that is a heavy price to pay for such a much artifact.

Issue #212, opened 2019-08-28
Images can become uneditble


dna113p:
In themcelan or themeclean flex, editing images can cause issues where you cannot scroll to offscreen components that have an image. The resize event never fires when images get loaded on page.

PR #222, opened 2019-09-10
Feature/resize observer polyfill

dna113p:
This branch contains bugfix/editview-scroll but I think this is a preferred way to add this polyfill so not to increase our bundle size. Fixes issue #212


reusr1:
@bkheadwire can you please check this PR if there are any side effects?

reusr1:
there seems to be an issue in IE11 - pages do not render the same (drag 2 text with media onto the page)

PR #223, opened 2019-09-12
Created servlet for listing tenants.

bkheadwire:


reusr1:
@bkheadwire can you change the path property to an object roots with a map for each type of root? 
something like
```
root: {
  sites: '/content/sites/themeclean',
 assets: '/content/assets/themeclean',
 templates;.. 
 objects: ...
}

Issue #97, opened 2019-06-19
Rich text editor link editing has not been styled


dna113p:
![image](https://user-images.githubusercontent.com/2624171/59779445-1f8dfb80-926d-11e9-8f65-ef88db93b55a.png)

dna113p:
Editing a link in the rich text editor should open up the path browser dialog that we already have built.

reusr1:
@dna113p I added the ability to override the modals from trumbowyg in the issues/97 branch - can you please hook this up to the url pathbrowser and the image browser?

bkheadwire:
Still has issues, see comments on the PR: https://github.com/headwirecom/peregrine-cms/pull/224

PR #224, opened 2019-09-12
#97: new modal for the richtext editor for links and images

reusr1:


reusr1:
@bkheadwire please review

bkheadwire:
There's a couple of issues:
1. Inserting a link by selecting a page in the content structure with the pathbrowser fails: https://youtu.be/s3R_cBx8Ua0. This appears to be related to the next issue...
2. Inserting a link using the link tab in the pathbrowser fails silently if you don't add a title.
3. Changing an existing link to a different path causes the text of the link to disappear. https://youtu.be/0skI6aerax8
4. If you insert an image and then try to choose a different image to replace it, the change is shown in the editor but not reflected in the content. Saving the edit does not persist the change. https://youtu.be/fG7rIQlLMig

dna113p:
@bkheadwire I have updated issues/97 and it should fix all of the things in your last comment.

reusr1:
@dna113p a link should always be inserted with .html at the end of the url and when editing a link the path browser should show the currently selected link. Same goes for editing an image

the path browser also has an issue with the footer - the path of the selected asset should fill the space on the left of the cancel/select buttons

bkheadwire:
All the issues I raised before are now fixed.

PR #227, opened 2019-09-16
translation for template component overlay and pages

FabianHaupt:


reusr1:
@FabianHaupt I believe this PR can be closed as the translations have been merged into develop with another PR already. Please close if you agree

FabianHaupt:
I agree that everything important is merged

Issue #195, opened 2019-08-21
Assets - Created field editable by pressing tab


NiklasFLuis:
<img width="302" alt="Bildschirmfoto 2019-08-21 um 12 08 32" src="https://user-images.githubusercontent.com/51320830/63423395-80bc8180-c40c-11e9-9db3-2a409bc5ba39.png">

The *Created* field of an asset is selectable by clicking into the Title-field and then pressing 'tab'. This will open the Date-picker and if you save a date it will be displayed in the Created-field.

<img width="306" alt="Bildschirmfoto 2019-08-21 um 12 11 06" src="https://user-images.githubusercontent.com/51320830/63423534-cbd69480-c40c-11e9-8c29-98c40c388773.png">
<img width="304" alt="Bildschirmfoto 2019-08-21 um 12 10 55" src="https://user-images.githubusercontent.com/51320830/63423544-cf6a1b80-c40c-11e9-81ab-479f6e59e0f8.png">

PR #232, opened 2019-09-23
#195: fix for datetime picker popping up modal when in preview mode

reusr1:


reusr1:
@bkheadwire please review and merge if ok

bkheadwire:
This does fix the date picker popping up. I made a slight change to the date format we use to serialize the dates to json since they were appearing as "invalid date" on the frontend.

There's an issue with assets not showing their creation dates but I think that's outside the scope of this bug so I'll open a separate ticket for that.

PR #237, opened 2019-09-25
Feature/actionrewrite

bkheadwire:


flx-sta:
@reusr1 as discussed: We agree to a merge 

but we have to write an issue to improve the promises that haven't been improved yet according to the "Anti-Pattern avoidance"

https://stackoverflow.com/questions/23803743/what-is-the-explicit-promise-construction-antipattern-and-how-do-i-avoid-it

PR #240, opened 2019-10-01
Pr/explorer dialog translation

flx-sta:
This PR provides a simple translation solution for the labels, placeholders and hints in the explorer-dialog.

It's the easiest solution to come up with for now but obviously not the best one.

_Did not test it throughly since we are currently having issue syncing our fork with the original repo. I apologize_


reusr1:
@Felix-Puetz this PR also contains part of the og:tag work from another PR - can you please clean this up so it's only related to translations? thank you

reusr1 (dismissed):
this PR contains other things than just translation (og:tag and word/character count). please remove

flx-sta:
@reusr1 I won't remove all translations, only the ones that are specifically for og:tags.
It doesn't hurt to have too many translations I'd say..

reusr1:
@Felix-Puetz was looking into this PR again - seems everything is pretty much merged from a translation perspective. can we close this PR?

flx-sta:
I would like this to be merged instead of closed since an important part of translation is covered here (not in a pretties way yet, but in a working way):

The explorer dialog translation implementation in the apiImpl.js

PR #241, opened 2019-10-01
Added disable/readonly to the button for the pathbrowser.

bkheadwire:


admin-base/ui.apps/src/main/content/jcr_root/apps/field/pathbrowser/template.vue
```diff
@@ -34,7 +34,7 @@
           :placeholder="schema.placeholder"
           :readonly="schema.readonly"
           @input="value = $event.target.value" />
-        <button v-on:click.stop.prevent="browse" class="btn-flat">
+        <button :disabled="disabled" v-on:click.stop.prevent="browse" class="btn-flat" v-bind:class="{hide: schema.readonly}">
```
reusr1: would it not be better to use a v-if on schema.readonly to make the html go away all together?

reusr1 (changes requested):

PR #242, opened 2019-10-01
Added our little "about" blurb from the welcome page to a modal acces…

bkheadwire:
…sible from the top nav.


reusr1:
@bkheadwire we need to support i18n in the about component

Issue #261, opened 2019-11-08
"Cannot read property 'experiences' of undefined" after first page load


flx-sta:
When opening the editor and trying to edit a component on the very first load of peregrine and the editor you can't press the "OK" Button because it's throwing an JS exception

<img width="606" alt="Screenshot 2019-11-08 at 08 18 46" src="https://user-images.githubusercontent.com/50131232/68457454-9fa70480-0200-11ea-9148-3f748a255c78.png">

flx-sta:
@reusr1 its just a quick fix. please merge asap

reusr1:
done

PR #262, opened 2019-11-08
fix for #261

flx-sta:
fix for "Cannot read property 'experiences' of undefined" #261


flx-sta:
Just a quick fix. Please merge asap

PR #276, opened 2019-11-27
adding help function

FloMarks:
With this PR the help function provided by peregrine is extended on several pages and new pages are equipped with it:
- Help function extended to "Home"
- Help feature extended to "Site"
- "Assets", "Templates", "Objects" and "Editor" equipped with help function

Currently, the help function is partly dependent on the content of each page. But there is already a ticket for us in which this dependency is to be resolved.


flx-sta:
@FloMarks I've changes this PR to "Ready for review" since it was still a draft

flx-sta (approved):
👍

PR #289, opened 2020-01-24
Merge actionrewrite into develop-sling11

flx-sta:


flx-sta:
I've manually checked all the usual usage stuff: looks good

PR #324, opened 2020-03-25
Inline Editors

reggie7:
The changes here introduce inline editing in a simple or rich text editor.
The new `percli` version is needed for it to work, so please update it first locally.
There is [a parallel PR](https://github.com/headwirecom/themeclean-flex/pull/89) for `themeclean-flex` that enables those 2 inline editors for a bunch of components. These PRs, if merged independently, should not break anything.


reusr1:
@reggie7 closing this PR as @Felix-Puetz implemented the inline editing differently

PR #366, opened 2020-04-14
Fixed many of the Peregrine Startup Exceptions including the Package …

schaefa:
…Manager exceptions from Composum


platform/base/core/src/main/java/com/peregrine/transform/operation/AbstractVipsImageTransformation.java
```diff
@@ -203,7 +203,7 @@ protected void transform0(ImageContext imageContext, String operationName, Strin
                         }
                         imageContext.resetImageStream(inputStream);
                     } catch (ExternalProcessException e) {
-                        log.error("Failed to execute VIPS", e);
+                        log.debug("Failed to execute VIPS command: " + operationName, e);
```
cmrockwell: Why the change to debug? Usually a log messages in catch clauses like this have error or warn level unless it is really expected and fully handled.

schaefa: This should not happen. Anyhow I changed it so that if not in debug mode it will just print a warning w/o a stack trace together with the operation name.

schaefa:
Manually merged into /issues/93

PR #368, opened 2020-04-15
Issues/367

cmrockwell:
There are a couple of error scenarios described in issues/367, which I would like to address. Thank you for considering merging these minor improvements to the UI stability.


reusr1:
@Felix-Puetz could you please review? than you

admin-base/ui.apps/src/main/js/apiImpl.js
```diff
@@ -693,7 +693,10 @@ class PerAdminImpl {
 
       updateWithForm('/admin/updateResource.json' + path + node.path, formData)
       // .then( (data) => this.populateNodesForBrowser(parentPath) )
-      .then(() => resolve())
+      .then(() => resolve()).catch( error => {
```
flx-sta: Put the `catch` into a new line please.
It has the same "priority" as `then`

flx-sta (changes requested):

admin-base/ui.apps/src/main/js/apiImpl.js
cmrockwell: done

flx-sta (approved):

PR #382, opened 2020-05-07
Default sitemap for the new structure

reggie7:
@reusr1 please let me know whether you want that PR split into refactor + main code.
The only thing needed for the sitemap to work is to add `domains` inside `templates/jcr:content`.


GastonGonzalez (approved):
This implementation looks good. My only comment is that we should either document the requirement to define a `domains` value for `templates/jcr:content` or my personal opinion is to capture the external domain during tenant creation. Also, post merge, the issues/93 branch will need a rebuild of the sling9 jar. These can happen post merge.

Issue #383, opened 2020-05-13
Tags component should show display value instead of the tag name


smcgrath0:
Please describe the problem.
Currently, the tags component shows the tags on the page by their names rather than their display values, which seems contrary to the purpose of having a display value in the first place.

Describe the solution you'd like
Display the display values for the tags instead of their names.

Additional context
https://user-images.githubusercontent.com/5474176/66245413-76a3c900-e6c2-11e9-9f47-57696cbddc70.PNG
https://user-images.githubusercontent.com/5474176/66245414-786d8c80-e6c2-11e9-9764-bc95b2208325.PNG
https://user-images.githubusercontent.com/5474176/66245417-7a375000-e6c2-11e9-9628-2a6a62cf8cf9.PNG

GastonGonzalez:
`issues/383` has been merged for both projects: peregrine-cms and themeclean-flex.  

<img width="1678" alt="Screen Shot 2020-06-02 at 11 53 00 AM" src="https://user-images.githubusercontent.com/1387966/83558974-b2e52e80-a4c8-11ea-8978-5bcc77b0f3ea.png">

PR #384, opened 2020-05-14
Issues/383

smcgrath0:
Thanks for contributing to Peregrine CMS!

**What does this implement/fix? Explain your changes.**

Tags component should show display value instead of the tag name

**Does this close any currently open issues?**

#383 

**Any other comments?**

**Where has this been tested?**
Local 
Screen Shot 2020-05-14 at 4 02 25 PM

*Brower (version):*
Chrome Mac

### DIFF
--- a/pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
+++ b/pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
@@ -336,7 +336,7 @@ public class PageModel extends Container {
             }
 
             if (this.value == null) {
-                this.value = this.name;
+                this.value = null;
             }
         }
 
--- a/pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
+++ b/pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
@@ -257,7 +257,6 @@ public class PageModel extends Container {
                 String tagString = tag.getValueMap().get("value", String.class);
                 Resource tagResource = tag.getResourceResolver().getResource(tagString);
                 if (tagResource != null) { 
-                // if (tag != null) {
                     answer.add(new Tag(tag));
                 }
             }
--- a/pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
+++ b/pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
@@ -322,10 +322,14 @@ public class PageModel extends Container {
             this.name = r.getName();
             String tag = r.getValueMap().get("value", String.class);
             Resource tagResource = r.getResourceResolver().getResource(tag);
-            this.value = tagResource.getValueMap().get("value", String.class);
+            if (tagResource != null) {
+                this.value = tagResource.getValueMap().get("value", String.class);
 
-            if (this.value == null) {
-                this.value = "Please Configure";
+                if (this.value == null) {
+                    this.value = this.name;
+                }
+            } else {
+                this.value = "";
             }
         }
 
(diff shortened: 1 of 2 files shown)


pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
```diff
@@ -320,7 +320,13 @@ public Tag(Resource r) {
             this.path = r.getPath();
             this.path = path.substring(path.indexOf("/jcr:content"));
             this.name = r.getName();
-            this.value = r.getValueMap().get("value", String.class);
+            String tag = r.getValueMap().get("value", String.class);
+            Resource tagResource = r.getResourceResolver().getResource(tag);
+            this.value = tagResource.getValueMap().get("value", String.class);
+
+            if (this.value == null) {
+                this.value = "Please Configure";
```
reusr1: I dont think please configure is a good value for no value - that part should be left to the UI to render (a fallback to the name when rendering the tag would be better)

pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
```diff
@@ -320,7 +320,13 @@ public Tag(Resource r) {
             this.path = r.getPath();
             this.path = path.substring(path.indexOf("/jcr:content"));
             this.name = r.getName();
-            this.value = r.getValueMap().get("value", String.class);
+            String tag = r.getValueMap().get("value", String.class);
+            Resource tagResource = r.getResourceResolver().getResource(tag);
+            this.value = tagResource.getValueMap().get("value", String.class);
+
```
reusr1, line 336: if you delete a tag after you added the tag to the page this code will fail with a null pointer exception.

pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
```diff
@@ -320,7 +320,13 @@ public Tag(Resource r) {
             this.path = r.getPath();
             this.path = path.substring(path.indexOf("/jcr:content"));
             this.name = r.getName();
```
reusr1, line 330: looks like this is a former error. the name of the tag should be fetched from the tag itself (same place as the value) and not from the name of the node that references it (probably fails to render the tag name correctly if the name of the tag ever changes)

reusr1 (changes requested):

pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
smcgrath0: That is correct, got objects working when deleted, can't use/see them. One other problem, cannot save/edit tags once placed on page. Think it has to do with our tags.infinity.json and how the data is stored, the tag objects are still there after deleting. 
Screen Shot 2020-05-15 at 1 23 53 PM

pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
```diff
@@ -254,7 +254,12 @@ public String getDataDefault() {
         List<Tag> answer = new ArrayList<Tag>();
         if(tags != null) {
             for(Resource tag: tags.getChildren()) {
-                answer.add(new Tag(tag));
+                String tagString = tag.getValueMap().get("value", String.class);
+                Resource tagResource = tag.getResourceResolver().getResource(tagString);
+                if (tagResource != null) { 
+                // if (tag != null) {
```
reusr1: no need to keep commented out code here

smcgrath0: Just pushed up without the comment

smcgrath0:
Should be good to go @reusr1

pagerenderer/vue/core/src/main/java/com/peregrine/pagerender/vue/models/PageModel.java
```diff
@@ -320,7 +328,15 @@ public Tag(Resource r) {
...
-            this.value = r.getValueMap().get("value", String.class);
+            String tag = r.getValueMap().get("value", String.class);
+            Resource tagResource = r.getResourceResolver().getResource(tag);
+            if (tagResource != null) {
+                this.value = tagResource.getValueMap().get("value", String.class);
+            }
+
+            if (this.value == null) {
+                this.value = null;
+            }
```
reusr1, line 339: not sure if this code really makes sense

Issue #390, opened 2020-05-18
assets reference list fails to render nicely


reusr1:
![image](https://user-images.githubusercontent.com/779992/82266641-12670a00-991f-11ea-91c5-1974b8432e81.png)

should render better, have tooltip on the url to see the full path and needs scrollbar if more pages use the same asset. It may be a good idea to use ... on the left side of the path as the urls usually differ the most to the right

GastonGonzalez:
This issue has been reviewed and merged. Per my comment in the PR, it may make sense to clean up the experience further and add support for collapsing duplicate references and rendering a count next to the referenced path.

![Screen Shot 2020-06-02 at 1 18 34 PM](https://user-images.githubusercontent.com/1387966/83566266-98b14d80-a4d4-11ea-83bc-856f226cd9cf.png)

PR #405, opened 2020-06-02
#390 fixed reference by list styles

smcgrath0:
fixes issue #390 
Screen Shot 2020-05-27 at 1 17 50 PM


GastonGonzalez (approved):
This change looks good. However, there are quite a bit of duplicate references. We may want to consider an enhancement to collapse the duplicates with a count. Something like this:

Screen Shot 2020-06-02 at 1 18 34 PM

Approving and merging this ticket as the interface has been cleaned up per the original issue.

Issue #380, opened 2020-04-27
og:image must be an absolute path


GastonGonzalez:
**Overview**

Open Graph images are not recognized by any social media providers since a relative URL is currently used by a Peregrine.

<img width="1277" alt="Screen Shot 2020-04-27 at 12 02 23 PM" src="https://user-images.githubusercontent.com/1387966/80410389-59616280-887f-11ea-94ef-2bf524b7b0d8.png">

<img width="1241" alt="Screen Shot 2020-04-27 at 12 03 10 PM" src="https://user-images.githubusercontent.com/1387966/80410411-5d8d8000-887f-11ea-9509-dbc8f5ed47b7.png">

**Example Page:** https://www.thetechprepper.com/posts/2020/04/ham-radio-go-kit.html

A potential solution is to generate an absolute URL using the first element under "Public Domain Names" on the root settings page.

<img width="1278" alt="Screen Shot 2020-04-27 at 12 08 07 PM" src="https://user-images.githubusercontent.com/1387966/80410641-c7a62500-887f-11ea-850a-66872566575c.png">

smcgrath0:
PR https://github.com/headwirecom/peregrine-cms/pull/410

PR #410, opened 2020-06-03
Issues/380

smcgrath0:
Fixes issue #380 

Regression Tested ✅ 

og:image now uses first public domain that is set in settings of the site, along with set og:image for that page. Will not display with domain is not set.
image


GastonGonzalez (changes requested):
@smcgrath0 - Please address the review comments related to your last commit. It was mostly around defense programming and logging.

smcgrath0:
image
og:image now works as intended, only first domain is used. Regression tested.

**Not sure if I'm suppose to add the property in these page models as well, pagerenderer/server/core/src/main/java/com/peregrine/pagerender/server/models/PageModel.java pagerenderer/server/core/src/main/java/com/peregrine/pagerender/react/models/PageModel.java

Tested with themeclean-flex, to test: 
#1 go to settings of the site your are currently on, go to 'configure root template'
#2 double click the parsys and scroll down to bottom dialog and add domain of your website
#3 go to the a page dialog and set the og:image property. 
#4 open that page and check with og chrome extension
image

GastonGonzalez (approved):

PR #450, opened 2020-07-15
Rewrite all content paths in all properties

GastonGonzalez:


admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
```diff
@@ -1732,12 +1732,24 @@ private void copyChild(Resource sourceChild, Resource targetParent, int depth) t
...
             // Copy grandchildren
             copyChildren(sourceChild, childTarget, depth + 1);
         }
 
+        private void replaceAllContentPathsInProps(final Resource resource) {
+            ValueMap properties = getModifiableProperties(resource, false);
+            Set<Entry<String, Object>> entries = properties.entrySet();
+            for (Entry<String, Object> entry : entries) {
+                final String curVal = (entry.getValue() instanceof String) ? ((String)entry.getValue()) : "";
+                if (StringUtils.isNotBlank(curVal) && curVal.contains("/content/" + fromName)) {
```
reusr1, line 1747: it looks like this may be a bit to wide of a contains statement. The issue was about A tags only

reusr1 (changes requested):

admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
reusr1: also not quite following why we would not do this before we create the node - we already go through all the properties a couple lines before your change: 

https://github.com/headwirecom/peregrine-cms/blob/develop/admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java#L1730

GastonGonzalez:
Closing this PR without merging. This PR will be used to address issues/444: https://github.com/headwirecom/peregrine-cms/pull/452

PR #457, opened 2020-07-18
Feature/firstlogin

reusr1:
- a basic implementation of an accept terms and conditions screen
- frontend implementation only, can be circumvented at this point


reusr1:
if we want to implement backend security for this feature as well we should do this by adding a group to the user when the terms and conditions are accepted.

GastonGonzalez (approved):
@reusr1 , outside of your "TODOs" this PR looks good. I corrected a couple of grammatical issues in the flow and added some error logging.

GastonGonzalez:
@reusr1 - A swagger update is needed given the new endpoint.

reusr1:
closing PR as this was merged into feature/tenantstate

Issue #460, opened 2020-07-20
avoid layout shifts by making sure we know the image width and height before rendering an image


reusr1:
- annotation for sling models to support fetching image width and height
- retrofit components to consume and render the additional information

schaefa:
I create a branch: issues/460-develop that contains the Image Info annotation and its handling as well as some Unit tests.

reusr1:
@schaefa would you mind taking some time to migrate this feature to develop-sling12? I ran a quick test but was not able to build this change against develop-sling12 due to reflection issues

```
[main] WARN org.apache.sling.models.impl.AdapterImplementations - Unable to reflect on com.peregrine.testmodels.DefaultImageInfoModel
java.lang.NoClassDefFoundError: org/apache/commons/lang/ClassUtils
	at org.apache.sling.models.impl.ReflectionUtil.mapPrimitiveClasses(ReflectionUtil.java:112)
	at org.apache.sling.models.impl.model.ModelClassConstructor.<init>(ModelClassConstructor.java:44)
	at org.apache.sling.models.impl.model.ModelClass.getConstructors(ModelClass.java:74)
	at org.apache.sling.models.impl.model.ModelClass.updateProcessorFactories(ModelClass.java:57)
	at org.apache.sling.models.impl.model.ModelClass.<init>(ModelClass.java:49)
	at org.apache.sling.models.impl.AdapterImplementations.addAll(AdapterImplementations.java:124)
	at org.apache.sling.models.impl.AdapterImplementations.addClassesAsAdapterAndImplementation(AdapterImplementations.java:110)
	at org.apache.sling.models.impl.ModelAdapterFactorySetup.createModelAdapterFactory(ModelAdapterFactorySetup.java:23)
	at com.peregrine.model.impl.ImageInfoInjectorTest.setup(ImageInfoInjectorTest.java:48)

```

ehdatheadwire:
@reusr1 and @reggie7 is this ticket still in progress or can it be closed? If required for 1.0.0 release, let's label it as such and make sure it's in the priority queue. Thank you!

reggie7:
I've worked on it and fixed it, I'd say yes. @reusr1 can you confirm?

reusr1:
@reggie7 @ehdatheadwire just tried this out, works well imho, closing the ticket

PR #462, opened 2020-07-20
#460 - Added the Image Info annotation to the Peregrine Sling Module

schaefa:
Fix for issue #460


platform/base/core/pom.xml
```diff
@@ -58,7 +58,9 @@
                             com.peregrine.replication,
                             com.peregrine.transform,
                             com.peregrine.transform.*,
-                            com.peregrine.commons.util
+                            com.peregrine.commons.util,
+                            com.peregrine.model.api,
+                            com.peregrine.model.impl
```
reusr1, line 63: do we really need to export the impl here?

platform/base/core/pom.xml
```diff
@@ -248,6 +242,10 @@
             <groupId>javax.inject</groupId>
             <artifactId>javax.inject</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
```
reusr1, line 246: do we need to rely on org.jetbrains for this? isnt @NotNull an existing annotation in java?

reusr1 (changes requested):
@schaefa can we remove the use of org.jetbrains here for the @NotNull annotation?

platform/base/core/pom.xml
reggie7: #671 handles that

platform/base/core/pom.xml
reggie7: #671 handles that

reggie7:
Replaced by #671

PR #465, opened 2020-07-22
issues/392

GastonGonzalez:


GastonGonzalez:
In general, the functionality for backing up, downloading, uploading and restoring a tenant site works. However, there are a few issues:

1. If you have two tenant users: `tenanta` and `tenantb` created by admin, any tenant can perform the following operations against any other tenant:
    * Read another tenant's backup metadata 
    * Download another tenant's backup

The permissions on the backups should be updated to only allow the tenant user to view/download their respective backups.

2. Add whitespace around site backup/restore buttons on the `/content/admin/pages/tenants/configure.html` page. They look a bit cluttered.

3. At some point, we should consider allowing multiple backups. The current implementation only allows for a single backup at a time. Update the UI and let the user know that only one backup is supported.

CC: @reusr1

PR #467, opened 2020-07-23
another (better) fix for the caret-jump to top issue (#420)

### LINKED ISSUE #420: Caret jump to the top (inline-editing)
How to reproduce:
1.  open a fresh page
2. focus into an inline-edit component and move the cursor to the end of it and press enter about 2 times to create new lines
3. reload the page
4. now click exactly into the same inline-edit component to be at the very end with the caret (right away!)
5. now type: cursor will jump to the top

flx-sta:


reusr1:
@Felix-Puetz I tried this together with #468 - unfortunately I see the cursor jump again when editing a rich text in a card

flx-sta:
true. 
Well come back to this issue later on since the current soltion works for now

PR #471, opened 2020-07-23
Issue/459

cmrockwell:
Hi
Please take a look at this PR to use repoinit for service users, ACL's and path creation.
Thanks 👍


GastonGonzalez (approved):
Thank you for the contribution. Keep up the great work. I ran full regression against `develop` with this change and it looks good.

PR #475, opened 2020-07-27
Issues/455

### LINKED ISSUE #455: inlineedit: toolbar stacks on smaller viewport sizes
image

we should make sure it stays as one line and once it does not fit anymore the overflow items should be in a dropdown as the last item of the toolbar

something similar how http://dojo.telerik.com/aBAPi moves overflow items into a dropdown

flx-sta:
rich-toolbar with collapse support for lower resolutions


flx-sta:
#455

flx-sta:
@reusr1 ready to merge

flx-sta:
merged into [feature/tenantstate](https://github.com/headwirecom/peregrine-cms/pull/468)

Issue #487, opened 2020-08-06
Fix stylesheets loading after vue styles


smcgrath0:
Stylesheets' styles should not override vue component styles. 

Issue can be visualized with multinavigation the component. Styles to hide the tab's list are being overwritten by themeflex styles to show them.
<img width="984" alt="Screen Shot 2020-08-06 at 3 59 24 PM" src="https://user-images.githubusercontent.com/49845255/89590936-d052b380-d7fd-11ea-9973-7f1d44a0391b.png">

reusr1:
closing this as we solved it differently in themecleanflex for now

PR #488, opened 2020-08-06
#487 fixed the browser interpretation of styles priority

smcgrath0:
Resolves issue: #487 

Stylesheets still preload, but now vue component styles will be placed after stylesheets in the html document. 

To test:
1. Drop multinavigation component on a page
2. give the component a root path of `/content/${your-site}/pages`
3. component should work normally as far as hovering the links


reusr1:
@smcgrath0 I am not sure about the approach you took here - are we sure the order the css is evaluated is deterministic this way (we dont know the order that will be used to 'enable' the css files)  - isnt the issue in general that the styles in https://github.com/headwirecom/themeclean-flex/blob/develop/fragments/navigation/template.vue should be either scoped locally or be unique enough not to have the conflict? Do we even need the styles in the component?

PR #508, opened 2020-08-13
Upgrade/sling12 pm java11

cmrockwell:
This Pull Request has the following goals

**Study Peregrine CMS on Sling 12**
Updated Peregrine CMS such that packages and bundles install and work properly on a Sling 12 starter.  The Sling 12 jar used in testing and development was created by cloning the [Sling Starter](https://github.com/apache/sling-org-apache-sling-starter) project, running `mvn clean install` and moving the resulting jar (target/org.apache.sling.starter-12-SNAPSHOT.jar) to the desired installation directory.  Project changes mainly involved versions of dependencies, and verifying that RepoInit scripts properly initialize Sling 12 for Peregrine CMS. Processes under the peregrine-cms/sling/peregrine-builder were not utilized.

**Studying Java 11**
Java 8 is basically end of service, so it's important to transition to the newer Java 11 (AdoptOpenJDK). 

**New Web Publishing Tab** 
The user interfaces for publishing (replicating) currently limited to the globe buttons. Publishing has enough complexity that a separate tab is warranted.

Screen Shot 2020-08-13 at 2 16 27 PM

In addition to the change to explorerpreviewcontent, the "Web Publishing" tab comprises two components: publishinginfo and publishingmodal. publishinginfo prints the publication status. When describing anything related to replication or activation in the user interface, it should be consistently called the same thing. Jargon words 'replication' and 'activation' are not user-friendly, which is why status is referred to as either Published, Unpublished, Stale or No Status. Stale indicates that a previously published page has since been changed, and therefore the published version is now "stale."

**New Web Publishing Modal** 
The UI seeks to bring the users into the know when it comes to supplemental resources such as images and templates. Equally important is to inform the user about downstream effects before they publish template changes. These data are represented in the user interface as "Uses" and "Used by" lists. 

Screen Shot 2020-08-13 at 1 47 32 PM

Each item in the "uses" and "used by" lists has a toggle. If the toggle is on, then that resource is included when publishing the main "Publish Item" 

The toggle will default to the "None" action if the resource is already published
image

The action is describe to the right of the toggle as either 
Publish (New) means the resource will be published for the first time
Publish (Again) means the resource has not been changed, but will be published again
Publish (Stale) means the resource has been changed, and stale content will be replaced
None  means the resource will not be published or unpublished 

**Tab Labels**
Adding text labels indicating the active tab is small but useful change. 
image

The following labels; Properties & Information, Open Graph Tags, References, Versioning, Web Publishing were used.

**Async Notice**
Web Publishing processes are started by the POST request to `repl.json` Upon successful initiation, the `replicate` API method then polls a new endpoint `listReplicationStatus.json` at an interval of 500ms. 

Screen Shot 2020-08-13 at 1 48 05 PM

It will check a maximum of 25 times over 12.5 seconds. If the processes did not succeed in this time, the user is notified that the action has "timed out"
Screen Shot 2020-08-07 at 4 29 05 PM

**Sling JUnit Tests Moved**
Sling JUnit Tests were moved to have a consolidated place for this style of JUnit testing. It makes use of the Example site for many tests. This PR also added testing for Remote replication, which could be useful for a production health check
http://localhost:8080/system/sling/junit/com.peregrine.slingjunit.RemoteReplAuthorJTest.html
However SlingJunit still requires LoginAdministrative.  The primary reason for Sling Junit is to have a fun way to do TDD within a developers Sling Instances. The Remote Replication tests added:
- authorRunmodes 
- authorDistributionServiceUserMappings,
- authorDistributionServiceUserAccess
- replicationMixin
- replicateOneAsset

**Refactoring and Minor bugs**
Minor bugs were fixed such as one that is currently reproducible on kestral-1. Clicking 'Objects' selecting an object and opening the References tab produces and JS error. The RefBy.json servlet was also called whenever something was selected in the explorer UI regardless of the active tab. This was refactored such that the servlet is called only when the Reference Tab is active. That fixes the JS bug as well as optimizing the network calls to `refBy.json`

The ReplicationServlet was moved to the admin-base/core which is where most other servlets are. This allows ReplicationServlet to depend on both platform and admin-base/core bundles without having circular dependencies. 

PerReplicable is a new adapter that intends to make getting and setting replication properties easies and with more consistency. It's usage was introduced to the `repl.json`, `ref.json`, `refBy.json` and `nodes.json` servlets.

Remote replication relies on Sling Content (Forward) Distribution. This Sling utility has few modes such as "ADD" and "DELETE". This limited set of operations leads to bugs when overwriting content. With a Stale content scenario (e.g.  content modifications are deletions or moves) when the system activates such changes. it recursively ADDs all the existing content to packages sent to the receiving instances using the ADD distribution type. Changes relating to deletions or moved content does not take effect. That results is duplications in the event of moved content (i.e. the publish instance shows it in the old and new locations). Therefore, this PR includes changes to `DistributionReplicationService` that deactivate resources before activation. This is simple solution solves the problem until SCD can be optimized for these publishing scenarios.


reggie7:
Hey, @GastonGonzalez has asked me to test the branch against [replication use cases](https://github.com/headwirecom/peregrine-cms/issues/534). Allow me to note my observations one-by-one in separate comments here as I go. They will come in irregular time intervals, I will note once I'm done testing.

reggie7:
# New tenants' resources published by default

## Description
When an author creates a new tenant/site from the same website theme that was previously replicated already - the resources created for it are already marked as published, even though they're actually not.

## Steps to Reproduce
1. Grab a new peregrine instance with e.g. themeclean-flex theme installed.
1. Create new website _A_.
1. Invoke **Site Setup Replication**.
1. Create new website _B_.
1. Open its **Pages**.
1. Check the _published_ icon.

**Expected Result**: Pages are marked as _unpublished_
**Actual Result**: Pages are marked as _published_.

## Analysis
**per:Replicated** and other similar properties are copied from the theme content to the new website instance.

PR #528, opened 2020-08-25
issues/473

GastonGonzalez:


platform/commons/src/main/java/com/peregrine/commons/util/PerUtil.java
```diff
@@ -914,4 +915,68 @@ public boolean doAdd(final Resource resource) {
...
     }
+
+    /**
+     * Extracts the tenant name from the resource
+     * @param resource a resource
+     * @return tenant name on success, and <code>null</code> otherwise
+     */
+    public static String getTenantNameFromResource(final Resource resource) {
+        return (resource != null)
+                ? getTenantNameFromPath(resource.getPath())
```
reusr1: this does not make sense to me. we can check the node names at the different levels directly

platform/commons/src/main/java/com/peregrine/commons/util/PerUtil.java
```diff
@@ -914,4 +915,68 @@ public boolean doAdd(final Resource resource) {
...
+                : null;
+    }
+
+    /**
+     * Extracts the tenant name for the path
+     * @param path a path
+     * @return tenant name on success, and <code>null</code> otherwise
+     */
+    public static String getTenantNameFromPath(final String path) {
```
reusr1, line 949: make sure `/content/iamatenant` also works - looking at the pattern this will not work right now

reusr1 (changes requested):

reusr1:
@GastonGonzalez forgot one thing here - did you test this with tenant users?

platform/commons/src/main/java/com/peregrine/commons/util/PerUtil.java
GastonGonzalez: I tested this use case on https://www.regexplanet.com/advanced/java/index.html. It works.

Screen Shot 2020-08-25 at 8 55 52 AM

GastonGonzalez:
> @GastonGonzalez forgot one thing here - did you test this with tenant users?

Yes, this works. The /content/tenant/var folder and temporary node are created and owned by the currently logged in tenant user. The old implementation throws an access exception since it was using /var.

platform/commons/src/main/java/com/peregrine/commons/util/PerUtil.java
GastonGonzalez: This method simply delegates to the `getTenantNameFromPath(String)` method. It's quite a bit simpler than iterating the nodes and checking node names., but I will update the implementation per your request.

GastonGonzalez: Method refactored. Please review.

PR #531, opened 2020-08-25
issues/529

GastonGonzalez:


GastonGonzalez:
@reusr1 - The only use case that fails, is copy by a tenant user. This will be resolved when this branch is merged. The `feature/tenantstate` branch has the backend code for copying to `/content/tenant/var`. I am happy to merge `feature/tenantstate` if you really need it in this branch now to test.

GastonGonzalez:
> @reusr1 - The only use case that fails, is copy by a tenant user. This will be resolved when this branch is merged. The `feature/tenantstate` branch has the backend code for copying to `/content/tenant/var`. I am happy to merge `feature/tenantstate` if you really need it in this branch now to test.

@reusr1 - I merged `feature/tenantstate` and tested the the tenant user write use case. We are good. This PR is all yours.

Issue #533, opened 2020-09-01
Site Map Scheduled Cache Re-builder is not invalidating the cache


GastonGonzalez:
**Issue:**
_Site Map Scheduled Cache Re-builder_ is not invalidating the cache for tenant sites.

**Branch:** feature/tenantstate

**Steps to Reproduce:**
1. Set the _Site Map Scheduled Cache Re-builder_ OSGi configuration to `0 45 * ? * * *`. This should run every hour at the 45 minute mark.

**Observed Results**
Cache is not invalidated and sitemap remains unchanged.

**Expected Results**
Cache is invalidated every hour and the sitemap is updated accordingly.

**Work Around:**
Manually delete `/var/sitemaps/files/site` and `/var/sitemaps/structure/site` in Composum when a sitemap update is needed.

Customer Reference: https://github.com/kestrel-1/report-an-issue/issues/20

reggie7:
Hey, I have checked it locally on the given branch. [com.peregrine.sitemap.impl.ScheduledSiteMapStructureCacheReBuilder](http://localhost:8080/system/console/configMgr/com.peregrine.sitemap.impl.ScheduledSiteMapStructureCacheReBuilder) defines a **Scheduler Expression** and at the given time the task calls [com.peregrine.sitemap.impl.ScheduledSiteMapStructureCacheReBuilder](http://localhost:8080/system/console/configMgr/com.peregrine.sitemap.impl.ScheduledSiteMapStructureCacheReBuilder) which has a **Debounce Interval** configuration. The actual page rebuilding will occur at the time given by CRON expression plus the debounce interval (by default it's 5m, please check and confirm in your environment). I believe this should call the rebuilding process at an equivalent of `0 50 * ? * * *`. Could you check that for me?

reggie7:
Additional issue noticed to get fixed:
1. Re-builder removes cache
2. It marks stuff to be rebuild
3. Debouncer starts counting time
4. Only when the debounce interval elapses - the cache gets back in place
The problem: in the meantime the cache dissapears.

GastonGonzalez:
@reggie7  - I reviewed this branch in dev and the cache invalidation works now. Before I merge this, here's what I'd like to see.

1. Please create a branch off `feature/tenantstate`  so that we have a cleaner merge.
2. Fix this new issue: The cache gets removed after the rebuilder marks stuff to get re-cached and before the debounce time elapses.
[...]

PR #541, opened 2020-09-09
fix cache rebuilder

reggie7:
Fix for [533](https://github.com/headwirecom/peregrine-cms/issues/533).


GastonGonzalez (approved):
The cache rebuilder is now working as designed. Approved.

PR #542, opened 2020-09-09
Implemented: Added support to localize content in JSON response using experiences

swapnilmmane:
A detailed explanation of the code and feature is available in this video.
https://youtu.be/KfDAPrCSgGs

Each node can manage the 'experiences' child node containing the language-specific localize node. Based on this return the localize content in JSON.
Please refer to the code for the sample node structure.

Thanks: @reusr1  for the inputs.

### DIFF
--- a/platform/base/core/src/main/java/com/peregrine/nodetypes/merge/PageMerge.java
+++ b/platform/base/core/src/main/java/com/peregrine/nodetypes/merge/PageMerge.java
@@ -171,8 +171,11 @@ public class PageMerge implements Use {
             if(val instanceof Map) {
                 Map map = (Map) val;
 
-                // Localize content using experiences.
-                localizeContent(map);
+                String xPerExperiences = request.getHeader("x-per-experiences");
+                if (!"off".equalsIgnoreCase(xPerExperiences)) {
+                    // Localize content using experiences.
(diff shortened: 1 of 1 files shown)


swapnilmmane:
We found some improvements and a new approach for the localization implementation.
Here are the two reasons, we thought of a new approach.

1. We compare the leading CMS (Storyblok, Contentful and others) and their process of allowing the user to localize the content. (they allow user to export the content so the user can localize it and re-import it)
2. The constant wish was to use the i18n library of Sling to localize the content https://sling.apache.org/documentation/bundles/internationalization-support-i18n.html

This gives birth to the new implementation approach
https://youtu.be/_0RGCJhwp-Y
which is scalable and widely adopted among other CMS.

So, now what we have done is, we used the i18n library of Sling to get localize content
And if user want to localize any content, they can export site data, localize it and import it again (currently the import is available form composum, in future, we can provide this option from admin panel)

Please have a look at this video and approach and let me know your thoughts.
Currently, I have not raised PR for new work in the official repo, but you can see the implementation here in these PRs.
https://github.com/hotwax/peregrine-cms/pull/42/files
https://github.com/hotwax/peregrine-cms/pull/43/files

Thanks!

PR #544, opened 2020-09-10
Allow only replicated pages to show up in a sitemap

reggie7:


GastonGonzalez (approved):
Confirmed desired behavior on dev-kestrel-1.dev environment.

Issue #534, opened 2020-09-01
Document replication use cases


GastonGonzalez:
The goal of this ticket is to identify all of the peregrine replication use cases. For now, let's ignore the actual replication implementation (i.e. Local FS Replication, Local Replication, Remote Replication, etc.) and focus on defining the use case on how replication should work.

1. Document use cases under the `docs/public/replication` folder.
2. Create `docs/public/replication/index.md`. This shall reference individual use case markdown files as a list.
3. Create one markdown file per use case (i.e. docs/public/replication/use-case-replicate-new-page.md)

```
# Replicate New Page

Created By: Gaston Gonzalez
Date Created:  1 September 2020
Date Updated: 1 September 2020
Actor(s): Peregrine CMS Author
Description: Author creates a new page and replicates it after initial editing. 

## Preconditions:
1. Page does not exist in system.

## Postconditions:
1. All assets referenced by the page are replicated along with the page. 

## Main Course:
1. Author creates new page
2. Author adds themeclean-flex components (heading and text)
3. Author replicates page

## Alternate Course:
1. ...

## Exceptions:
1. ...
```

Please use the above template as a reference. Please come up with a list of other use cases. Some that come to mind are:
* Replication status icon changes on replication
* Replicate updated page
* Replicate root site
* Replicate asset
* Replicate object
* Replicate template
* Replicate sitemap
* Template updated
...

GastonGonzalez:
@reggie7 - Please add a use case that covers how the sitemap is updated upon replication. My opinion is that we should only include pages in the sitemap that have been replicated.

reggie7:
@GastonGonzalez can we close?

GastonGonzalez:
Closing ticket.

PR #562, opened 2020-09-18
replication use cases

reggie7:
[534](https://github.com/headwirecom/peregrine-cms/issues/534)


GastonGonzalez:
@reggie7, This is a really good start, but it looks like these use cases mainly deal with happy path use cases. I believe there may be quite a few edge use cases that may exist. For example, do we really want to to remove all pages when a template is removed? Think large site deployments.

Cris added support for versioning. Could we leverage the versioning support to provide a better a replication experience for users? Just a thought, if we consider versioning along with replication, we could likely remove a template, but still allow all pages in the system that use it to still be available since it would be tied to a particular version. In this model, new pages would not be able to use the template, but previously created pages would continue to work. Likewise, template changes that can seriously break a page (i.e. the removal of container) would still work. Old pages could still leverage the template prior to this change and new pages could use the newly updated template.

Give this some thought and double check the uses cases for missing edge cases like this.

Our goal here is really think about how users will use Peregrine's replication system in a production setting. Also, as you re-review these uses cases, you may want to think about some challenges that may come up based on the Local Replication Service (node-to-node) and Local FS Replication Service. 

If you really want to get these use cases fully documented with the edge cases, consider how you would "break" the customer facing site through replication.Try this use both Local Replication Service (node-to-node) and Local FS Replication Service.

cmrockwell:
I'm reviewing this PR wrt issue #534 and a bit confused. This issue talks about setting more definition for replication processes, which is essentially a specification task. A very important and useful task. However, really good specifications or user stories for replication use cases still appear lacking, so what is getting implemented by this PR? 

Looking at the files changed, it is hard to tell
https://github.com/headwirecom/peregrine-cms/pull/562/files 

feature/tenantstate was merged to the develop branches after extensive teamwork to resolve merge conflicts. When I see massive amount of files changed in a PR dealing with feature/tenantstate, I ask, it is happening again? 

I would respectfully suggest applying the changes to a new branch (taken from develop-sling12), and then writing a new PR, so it becomes clearer from the outside what is being done.

reggie7:
Nice catch @cmrockwell - I've pushed an updated version of the PR branch with `develop-sling12` merged into it ~2h ago, but didn't realize the base branch here is still set to `feature/tenantstate`. It's changed now already, so the diff should be clear.

At the moment I'm still working to supply extended descriptions of the replication cases as requested in [the source ticket](https://github.com/headwirecom/peregrine-cms/issues/534) and [the above comment](https://github.com/headwirecom/peregrine-cms/pull/562#issuecomment-698637948). Feel free to look at the 4 **Root Site** cases if you have the time to review this PR - this is what I've recently updated. I'll push other updates soon, hopefully you'll get notification emails about them.

Any suggestions will be welcome as to what edge cases we might encounter within the replication processes, since you've worked in this area, your input will be invaluable.

docs/public/replication/asset-deleted.md
```diff
@@ -0,0 +1,30 @@
...
+## Postconditions:
+1. The Asset is now removed from replication cache.
+1. All unreferenced tags associated with the Asset
+   are now removed from replication cache.
+
+## Main Course:
+1. Author deletes an Asset.
+1. System presents the author with a confirmation dialog asking about removal of
+   replicated resources.
+1. System presents the author a list of other pages that reference the Asset to be removed.
```
cmrockwell: There may be a gap between what is said and what is meant by this line. Seems to say that a list pages, which use the asset, will be removed. Perhaps what is meant is asset references will be removed from a list of pages.

GastonGonzalez: @cmrockwell - You are correct. The pages will NOT be removed. The question is really this: 
1. Should we actually remove the references from the various nodes?
2. Or, should we simply inform the user that the list of pages will be impacted if this action is executed?

CC: @reusr1

docs/public/replication/index.md
```diff
@@ -0,0 +1,84 @@
...
+![Replication Strategies](replication-types.jpg)
+
+Repository is quite safe in most cases since the pages can be rendered on the fly.  
+Much more attention has to come to the cache-like replication options, i.e. those, where we store
+final renderings of the replicated resources. A lot can go wrong if we don't handle the references
+properly.
+
+## Versioning Role
+We should version the published resources and use those versions to render published resources that depend on them.
+Let's not though, that it will be much more feasible to simply keep an up-to-date published tree of resources
```
GastonGonzalez: Typo: "Let's not" -> "Let's note"

GastonGonzalez:
@reggie7 - Please add a use case for the sitemap. We need to ensure that the sitemap regardless of implementation reflects the current publication state. Thanks. This is top of mind for @reusr1 and will likely be one of the first issues we tackle. Thanks.

reggie7:
Sitemap use cases are restored now, please check @GastonGonzalez - they're quite simplistic since I assume no issues with them - it's basically about writing and removing text files based on existing cache.

reggie7:
Well, we trigger the sitemap update also on resource changes, so a page change does start the update process. There will only be a short time delay coming from debounce interval (think couple of seconds). We just need to create simple hooks that will implement certain interface. That's from code perspective. But I assume we want to have that requirement explicitly added in the use case?

PR #571, opened 2020-09-24
Allow read on index.html for all_tenants

reggie7:
I was also wondering whether an alternative solution could be to set `sling:redirect` on `jcr:root` just as the `index.html`'s `js` does, but for now let's leave it as is. Unless you guys prefer the above - then just let me know.
Anyways - this will not work on a fresh instance right away, a restart is needed. I assume this is another issue to be handled separately (`org.apache.sling.jcr.repoinit.RepositoryInitializer-peregrine.config` seems to kick in only after 2+ instance start) - when fixed, it will propagate here as well.


reusr1:
@reggie7 can you also add this type of rule for favicon.ico and the robots.txt file?

GastonGonzalez (approved):

GastonGonzalez:
@reggie7 and @reusr1 . I already merged this before seeing your comment about allowing `/robots.txt` and `/favicon.ico` for the admin. I made this change directly on develop-sling12 with this comment: `https://github.com/headwirecom/peregrine-cms/commit/0c2988d9120a0fce956538e9545d48140569f096`

PR #572, opened 2020-09-25
Updated Sling startup to use a Feature Archive

GastonGonzalez:


docker/Dockerfile
```diff
@@ -24,6 +24,9 @@ ENV PATH "/usr/local/lib/nodejs/node-v10.16.0-linux-x64/bin:$PATH"
 COPY files/* /app/binaries/
 COPY scripts/*.sh /app/scripts/
 
+RUN mkdir -p /root/.m2
+COPY overlay/root/.m2/settings.xml /root/.m2/
```
reusr1, line 28: is this really needed? now that the settings.xml file you overlay has the profile disabled - why would we need to copy a settings.xml in this case and if we do, would it not just be easier to have settings.xml without the profile altogether?

reusr1 (changes requested):

docker/Dockerfile
GastonGonzalez: It was added to support a more customizable maven build environment. While the profile is no longer active by default, I believe the plumbing still makes sense for potential future use. There's no harm in leaving this hook here as it has already been tested IMHO.

PR #573, opened 2020-09-26
Implemented: Added Adobe Stock integration in Peregrine.

tsaurabh74:


reusr1:
@tsaurabh74 thank you so much for this PR! Would you mind adding a small .md file to /docs and explain the feature (how does it need to be configured). I saw in the code somewhere that it needs an API KEY and application name to make the requests to Adobe Stock. We probably should think about how we configure this setting - is this a per user, per website or per server setting?

flx-sta (commented):
Looks good. Just gotta find a way to get the api key support

PR #584, opened 2020-10-06
Sling12: issues/581 - use title instead of name in admin ui

### LINKED ISSUE #581: use title instead of name in admin UI
A couple of areas in the admin UI default to the node name for display instead of the title (such as the display of the current tenant or the page navigation while editing a page. Please go through the UI and make sure we always display the title if available instead of the node name (node name should be fallback).

flx-sta:
PR for `develop-sling12` branch


GastonGonzalez (approved):
Local testing looks good. Confirmed friendly title on tenant page and on dropdown while editing page.

Issue #577, opened 2020-10-02
Prevent delete operation if published


GastonGonzalez:
The goal of this ticket is to ensure that Peregrine does not allow users to delete a resource if it's in a published state. This should be enforced in the admin UI as well as the API. This change should help mitigate state issues related to replication.

**Use Case**
TODO: @reggie7 - please consider the impact and changes needed for delete|move|copy|rename|...

reggie7:
Looking at how the structures will be affected - _move_/_rename_ will be painful to implement :( Both of these operations should probably not be blocked for the author by the replication status. Tha's my gut feeling.
_Copy_ only requires us to skip the replication properties.

reusr1:
@reggie7 @GastonGonzalez @cmrockwell it feels to me like pages probably should not be deletable/movable once published until they unpublished again?

GastonGonzalez:
> @reggie7 @GastonGonzalez @cmrockwell it feels to me like pages probably should not be deletable/movable once published until they unpublished again?

@cmrockwell I agree. Anything that puts the published cache or publish/live server in a stale/invalid state should be prevented. We'd like to handle the move/rename in a separate ticket as a fast follower.

GastonGonzalez:
This issue has been resolved. A new issue will be logged to provide similar support for move and copy operations.

PR #585, opened 2020-10-07
Prevent delete operation if published

reggie7:
[577](https://github.com/headwirecom/peregrine-cms/issues/577) - I'm leaving the action button in order to be transparent for the user - this way they will know what is going on:

delete

Please note that I've set java to `11` in the main `pom.xml`. Let me knoe what you think.


GastonGonzalez:
@reggie7 - I really like what I am seeing. I'm still testing and noticed one edge case so far that does not work. I believe it is a reference issue and technically not an issue with this PR.

Use Case
1. Create a page (p1) and publish it.
2. Create a second page (p2). Add a card component and create a card that links to (p1). Publish p2.
3. Unpublish p1.
4. Delete p1.

Observed Behavior
1. p1 is allowed to be deleted.

Expected Behavior
1. p1 is referenced in p2. It should not be allowed for deletion.

GastonGonzalez:
@reggie7 - This PR is very difficult to determine the exact changes due to the number of clean up updates. Can you please remove the clean up updates. We need to avoid refactoring in these types of tickets. Refactoring should be done in a separate ticket. Also, please do not use wildcard imports.

GastonGonzalez:
@reggie7 - This change has a performance impact. https://github.com/headwirecom/peregrine-cms/pull/585/files#diff-2ca240fef4c0c7351fe4a53bd9a3ee13R376. This looks like another refactoring. It should be reverted.

GastonGonzalez:
@reggie7 - What would the effort be to change the toast popup when the delete button is clicked on a published resource to prompt a modal with the same message as well as show the references?

GastonGonzalez:
@reggie7 - The page move is also a problem. I think this can be a separate ticket.

Use Case
1. Create page (p1) and publish
2. Move p1 to a new location

Observed Behavior
1. p1 is moved but published content is not left on cache or publish/live server

Expected Behavior
1. We need to message the user that this page is published and needs to be unpublished to support the move. A broader conversion on Slack is likely needed.

GastonGonzalez:
> @reggie7 - What would the effort be to change the toast popup when the delete button is clicked on a published resource to prompt a modal with the same message as well as show the references?

@reggie7 - Let's make this PR simple. Let's just keep the toast message/functionality as-is.

GastonGonzalez:
> @reggie7 - The page move is also a problem. I think this can be a separate ticket.
> 
> Use Case
> 
> 1. Create page (p1) and publish
> 2. Move p1 to a new location
> 
> Observed Behavior
> 
> 1. p1 is moved but published content is not left on cache or publish/live server
> 
> Expected Behavior
> 
> 1. We need to message the user that this page is published and needs to be unpublished to support the move. A broader conversion on Slack is likely needed.

@reggie7 - We will create another ticket for this. We'll keep it simple in the new ticket. Per @reusr1 , we will just prevent any behaviors from being performed that can lead to an unstable state (i.e. move and rename).

reggie7:
> I really like what I am seeing. I'm still testing and noticed one edge case so far that does not work. I believe it is a reference issue and technically not an issue with this PR.

Hmm, true, haven't considered this scenario here. I say we make a separate ticket for this too, what do you think? Or want me to rather include it here already?

reggie7:
All other comments are taken care of (or will be in separate tickets). Looks much cleaner now indeed. Just please use diff with whitespace comparison off - I see there's a file with line endings changed from Windows to Unix style, let's keep that in.

GastonGonzalez:
> > I really like what I am seeing. I'm still testing and noticed one edge case so far that does not work. I believe it is a reference issue and technically not an issue with this PR.
> 
> Hmm, true, haven't considered this scenario here. I say we make a separate ticket for this too, what do you think? Or want me to rather include it here already?

@reggie7 - Please create a separate ticket. I would like to limit the number of changes in this PR.

admin-base/core/src/main/java/com/peregrine/admin/servlets/DeletePageServlet.java
```diff
@@ -75,31 +78,41 @@
 
     public static final String FAILED_TO_DELETE_NODE = "Failed to delete node: ";
 
-    @Reference
-    ModelFactory modelFactory;
```
GastonGonzalez, line 79: @reggie7  - Was the ModelFactory not used in this class?

GastonGonzalez:
@reggie7 - I retested your new branch: `issues/577-prevent-delete-operation-if-published-simplified`. It looks good. Please create a new PR off the new branch. I will close this PR without merging it.

GastonGonzalez (approved):

PR #594, opened 2020-10-12
develop-sling12: remove ablity to drop into non-droptarget tempalte component (#591)

### LINKED ISSUE #591: breaking rendering of pages through dropping component on page
steps to reproduce: 

- create a site with themecleanflex
- create a page based on the article skeleton page
- drop a rich text component onto the 'related content' tile

result: page rendering is broken

flx-sta:


reusr1:
@Felix-Puetz while this prevents a user from dropping a component onto the non container components from the template in the editor I can unfortunately not drop a component in the right rail area below the related content in a page created with the article skeleton page. I see the selection cursor quickly show up and hide again. Seems to be still buggy

flx-sta:
@reusr1 ill look into it and see what i can do..

flx-sta:
@reusr1 please have a new look at it.
There is still one issue that everything dropped with drop below the existing "related content" but im not quite sure if this is supposed to happen or a bug.
Trying to "prevent" this visual behaviour would be kinda tricky and a very specific hard-coded error handling

flx-sta:
poke

reusr1:
@Felix-Puetz tnak you, looks good to me

PR #601, opened 2020-10-14
prevent move and rename operations if published

reggie7:


GastonGonzalez (approved):
@reggie7, this PR looks good. I ran full regression as well as tested the delete, move and rename UI interactions for both published and unpublished states. No issues found.

PR #628, opened 2020-10-23
Fix create tenant properties copies

reggie7:


GastonGonzalez (changes requested):
@reggie7 - I am seeing a high number of exceptions in logs/peregrine.cms.log when a site is replicated. If these are expected, we should change the log level.

GastonGonzalez (approved):

PR #634, opened 2020-10-26
Sling12/inline edit issues

### LINKED ISSUE #637: image gets removed after changing dimensions
### Description
On fresh tenant/website and fresh pages images are delted upon editing their dimensions

### Reproduction
* log into pcms
* Create a new Website
  * press"create" and the end (and not "create and edit!")
* Go to "Pages"
* Create new site 
  * press "create and edit!" at the end (and not "create")
* click twice into this rich-text
  * image
* insert the image `/icons/launcher-icon-1x.png` 
* doubleclick the freshly inserted icon to open the modal again
* switch to the link-tab image 
* change the width and height (e.g. 500, 200)
* click on "select"
* BOOM! image is gone

flx-sta:
### FOR DEVELOP-SLING12

I've reviewed some issues with related to inline-editing.

the `icon` action wasn't showing up in the toolbar anymore.
Turned out that `populateIcons:/state/tenant` was missing as a loader inside `admin/pages/pages/edit`

### DIFF
--- a/admin-base/ui.apps/src/main/content/jcr_root/content/admin/pages/pages/edit/.content.xml
+++ b/admin-base/ui.apps/src/main/content/jcr_root/content/admin/pages/pages/edit/.content.xml
@@ -37,7 +37,7 @@
     dataFrom="/pageView/path"
     dataDefault="/content/admin/pages"
     suffixToParameter="[path,/pageView/path,path,/state/tools/page]"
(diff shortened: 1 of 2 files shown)


flx-sta:
poke

reusr1:
@Felix-Puetz I tried this one - the insert icon button shows up again but the icon is only inserted as a placeholder and the placeholder is not turned into an icon. Was that how it worked before? 
Screenshot from 2020-11-23 11-00-29

flx-sta:
@reusr1 yes that was the original behaviour which was agreed too.
First they were added as `` but that was not the wanted solution and we postponed the proper solution and used this placeholder strategie

PR #642, opened 2020-11-04
make use of multi stage docker builds

reusr1:
this change to the docker file into a multistage docker build and removing the git repos towards the end of the build before creating the final image drops the final image size from ~4.5GB to 1.25GB

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM adoptopenjdk/openjdk11:latest
+FROM adoptopenjdk/openjdk11:latest as build
 
 ARG PEREGRINECMS_REPO=https://github.com/headwirecom/peregrine-cms.git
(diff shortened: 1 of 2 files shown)


GastonGonzalez (approved):
@reusr1 , this is a great PR. The [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) support in Docker is a great move forward to eliminate the build-time dependencies from the final image. The time for cloning `peregrine-cms` is still problematic. Should we update the `git clone` to only pull the branch that is needed like we did for the Toolbox project?

reusr1:
@GastonGonzalez sure - I think that would be a good idea

GastonGonzalez:
> @GastonGonzalez sure - I think that would be a good idea

Great. I'll add that to my list of todos for tomorrow. This is a quick fix.

GastonGonzalez:
@reusr1 - Please review at your convenience the last commit

PR #653, opened 2020-11-17
Remove unused imports

reggie7:


platform/base/core/src/main/java/com/peregrine/adaption/impl/PeregrineAdapterFactory.java
```diff
@@ -38,11 +38,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.jcr.Node;
-import javax.jcr.RepositoryException;
-
 import static com.peregrine.commons.util.PerConstants.*;
-import static com.peregrine.commons.util.PerConstants.PER_REPLICATION;
```
reusr1, line 45: we probably should remove the line before instead

reggie7: Hmm, good catch, though I suggest this instead: let's keep the refactoring PRs a single-topic ones. Maybe I deal with `.*` imports in a separate one? I hear that the team prefers them expanded, so we can expand them all in another PR. Can we do it this way?

flx-sta (approved):
i dont know everything of the backend but looks fine for me
image
image

cmrockwell (approved):
Looks okay to me

PR #664, opened 2020-11-25
sling12/Issues/448: make sure we dont have large external dependencies

flx-sta:
pr redo after messing up #656


Felix-Puetz:
Review
@reusr1 
### After Login
image
* I don't think we should be loading images of the CMS itself from an external source (headwire.com)
* We should use local versions of the css for `font-awesome` and `ionicons`

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/toolingpage/toolingpage.html
```diff
@@ -113,7 +113,6 @@
 </sly>
 
 <sly data-sly-include="renderer.html"></sly>
-<script src="https://unpkg.com/vue-codemirror-lite/dist/vuecodemirror.min.js"></script>
```
flx-sta, line 116: This worked 👍

flx-sta (commented):
@reusr1 
### After Login
image
* I don't think we should be loading images of the CMS itself from an external source (headwire.com)
* We should use local versions of the css for `font-awesome` and `ionicons`
otherwise offline the rich-toolbar looks like this
image


### On Edit Page
image
* You did mention this though so I'll leave it to you if you want to keep it like this or want to download these too

flx-sta:
~~@reusr1 please review the changes i made since i can't review it myself... except waiting for a new docker image~~

nevermind it is broken

flx-sta:
font-awesome whole module is now exported into the felibs directory due to dependencies of fonts

PR #667, opened 2020-11-27
selection is now cleared upon changing components (#639)

### LINKED ISSUE #639: text-selection not cleared upon selecting different component
When editing a page and you select a part of text inside one component this selection is not cleared upon changing into another one resulting in unwanted behaviour upon using the richtoolbar

flx-sta:
End-to-End-Test can be found here
https://github.com/Felix-Puetz/pcms-testing/blob/1baca8e9f2b761d8b847e29c8bb13954079cdc3c/src/tests/inline-editing.js#L35
https://github.com/Felix-Puetz/pcms-testing/blob/1baca8e9f2b761d8b847e29c8bb13954079cdc3c/src/tests/editor-panel-editing.js#L36

### DIFF
--- a/admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/contentview/template.vue
+++ b/admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/contentview/template.vue
@@ -406,15 +406,19 @@ export default {
               }).then(() => {
                 $perAdminApp.action(vm, 'showComponentEdit', vm.path).then(() => {
                   vm.flushInlineState()
-                  vm.$nextTick(vm.pingToolbar)
-                  vm.$nextTick(focusElement(this.target))
+                  vm.$nextTick(() => {
+                    vm.pingToolbar()
+                    focusElement(this.target, this.iframe.win)
+                  })
                 })
               })
             } else {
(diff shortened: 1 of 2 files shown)


reusr1:
@Felix-Puetz lets look a this PR in a sharing session - I am not sure if I like some of the ways this affects overall editing

flx-sta:
Okay.
But just to be clear the behaviour is generally the requested one, but you just aren't sure anymore if you want to have this behaviour ?

reggie7 (approved):

flx-sta:
fixed the issue of not being able to tab through anymore

reusr1:
@Felix-Puetz I still see issues with this PR: 

- on the homepage teaser for the demo sites: click into content editable for title, click into title field in dialog, add text to dialog (text is not added in content editable) change text in subtitle (text is changed on page)

flx-sta:
> @Felix-Puetz I still see issues with this PR:
> 
> * on the homepage teaser for the demo sites: click into content editable for title, click into title field in dialog, add text to dialog (text is not added in content editable) change text in subtitle (text is changed on page)

as discussed will this address a new issue
#679

PR #670, opened 2020-12-02
Fix inconsistent UX of delete operation

### LINKED ISSUE #602: Inconsistent UX of Delete operation
We would need to use the same confirmation dialods for both delete buttons. Currently the one under **Actions**

delete

uses

confirm

while I assume that the proper one would be

proper

reggie7:
Fix for #602


admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/explorer/template.vue
```diff
@@ -653,27 +648,34 @@ export default {
...
+                            resolve()
                         }
-                    }
+                    })
                 })
             },
 
+            deletePage: function(me, target) {
+                const { resourceType, path } = target
+                let action = 'folder'
```
flx-sta: I thinks it's a bit confusing here that `action` is the name of the variable when it is passed as `type` to `handleDelete()`

flx-sta (changes requested):
Only the small `action -> Type` change please

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/explorer/template.vue
reggie7: Thanks, good catch!

flx-sta (approved):

flx-sta:
I've wrote an end-to-end test for this since it's an interessting scenario for the page-object pattern.
Because `PagesPage`, `ObjectsPage`, `AssetsPage` and `TemplatesPage` would be very similar and to prevent this class extension is used

https://github.com/Felix-Puetz/pcms-testing/pull/2

reusr1 (approved):

Issue #17, opened 2019-02-19
Dragging component in editor throws server error


devintuffy:
To replicate:
- Go to Pages, click on Editor for a page (for this I used the built-in "Write" page)
- Press and hold on a draggable component, and attempt to move it to a new location on the page

Result: Component is not moved, and the server responds with a 400 error. Console output:

```
POST http://localhost:8080/perapi/admin/moveNodeTo.json/content/sites/docs/write/jcr:content/nav 400 (Bad Request)[error]
[apiImpl]                  Update with Form request to undefined failed
Uncaught (in promise) Error: Request failed with status code 400
    at e.exports (spread.js:25)
    at e.exports (spread.js:25)
    at XMLHttpRequest.l.(anonymous function) 

```
And here's the server response:

```
{
  "message" : "To Move a Node the Reference Resource must be provided",
  "exception" : [
    "com.peregrine.admin.resource.AdminResourceHandler$ManagementException: To Move a Node the Reference Resource must be provided",
    "\tat com.peregrine.admin.resource.AdminResourceHandlerService.moveNode(AdminResourceHandlerService.java:331)",
    "\tat com.peregrine.admin.servlets.MoveNodeTo.handleRequest(MoveNodeTo.java:105)",
    "\tat com.peregrine.commons.servlets.AbstractBaseServlet.doRequest(AbstractBaseServlet.java:78)",
    "\tat com.peregrine.commons.servlets.AbstractBaseServlet.doPost(AbstractBaseServlet.java:72)",
    "\tat org.apache.sling.api.servlets.SlingAllMethodsServlet.mayService(SlingAllMethodsServlet.java:149)",
    "\tat org.apache.sling.api.servlets.SlingSafeMethodsServlet.service(SlingSafeMethodsServlet.java:346)",
    "\tat org.apache.sling.api.servlets.SlingSafeMethodsServlet.service(SlingSafeMethodsServlet.java:378)",
    "\tat org.apache.sling.engine.impl.request.RequestData.service(RequestData.java:552)",
    "\tat org.apache.sling.engine.impl.filter.SlingComponentFilterChain.render(SlingComponentFilterChain.java:44)",
    "\tat org.apache.sling.engine.impl.filter.AbstractSlingFilterChain.doFilter(AbstractSlingFilterChain.java:77)",
    "\tat org.apache.sling.engine.impl.SlingRequestProcessorImpl.processComponent(SlingRequestProcessorImpl.java:282)",
[...]

PR #671, opened 2020-12-03
#460 - Added the Image Info annotation to the Peregrine Sling Module

reggie7:
Implements #460.
Replaces #462.
Base for [percli#24](https://github.com/headwirecom/percli/pull/24) and [themeclean-flex#175](https://github.com/headwirecom/themeclean-flex/pull/175).


platform/base/core/src/main/java/com/peregrine/model/impl/ImageInfoInjector.java
```diff
@@ -0,0 +1,117 @@
...
+        if (isNull(annotation)) {
+            return null;
+        }
+
+        final BufferedImage image = getImage(resource, ((ImageInfo) annotation).name());
+        if (isNull(image)) {
+            return null;
+        }
+
+        return new Dimension(image.getWidth(), image.getHeight());
```
reusr1: I am not sure if I like this approach - we fetch the image and parse it - that is a lot of computational overhead for each image served. We already extract the image width and height into the meta data at upload of the image. Would it not make more sense to use the data from there?  @schaefa @reggie7 ?

reusr1 (dismissed):

platform/base/core/src/main/java/com/peregrine/model/impl/ImageInfoInjector.java
reggie7: Yup, it does, I have actually found it during my work on #17, adjusted and just pushed the code a couple of minutes ago.

PR #676, opened 2020-12-07
Issues/460 develop sling12 junit

cmrockwell:
Created tests for image dimensions. SVG is not yet supported. Assets could be document formats or many other formats, so mine-type tests before setting dimensions could be a good thing.


reggie7:
Thanks @cmrockwell, looks good initially, will look and, most likely, merge tomorrow.

PR #686, opened 2020-12-11
Issues/460 develop xmlhelper for svg

cmrockwell:
PerAssetImpl appeared to be getting a little busy with all the SVG String operations. And since SVG is XML, probably would be better to use Java libraries for reading the doc and attributes instead of parsing Strings directly from input streams. I moved some of the methods that can static into XMLHelper, and looked through to try and close inputstreams and ImageInputStream

### DIFF
--- a/platform/base/core/src/main/java/com/peregrine/adaption/PerAsset.java
+++ b/platform/base/core/src/main/java/com/peregrine/adaption/PerAsset.java
@@ -119,5 +119,11 @@ public interface PerAsset
      * get dimension (or null) for asset
      */
     Dimension getDimension() throws RepositoryException, IOException;
-    
+
+
+    /**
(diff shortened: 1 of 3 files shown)


reggie7:
@cmrockwell I was on your side in [the water-cooler](meet.google.com/kcu-qjdc-kcn) meeting regarding code changes to use `xml` classes for the job of extracting `svg` size. But I have just run the branch and the time to upload an `svg` file grew from ~**50 ms** to ~**16 s**, so it multiplied 320 times (my localhost, 4 small test files). I have to side with @reusr1 on the performance issue here. I'm still fine with the general approach and code changes from readability perspective, but we'd need to go back to the same order of magnitude regarding performance.

cmrockwell:
Alright.

Issue #494, opened 2020-08-12
better range slider for forms 


reusr1:
admin-base/ui.apps/src/main/content/jcr_root/apps/field/material-range currently renders a range slider (with min/max values). This however does not cover all cases of the input data and is a bit cumbersome to set exact values. 

we should make sure we can:
- deal with the case where we have no data for the field and no default
- be able to see the current value of the range without clicking it
- be able to enter an exact value
- remove the value

reusr1:
current work in https://github.com/headwirecom/peregrine-cms/tree/issues/494

reusr1:
@Felix-Puetz assigned this issue to you - let's see if we can create a better UI for the slider

PR #688, opened 2020-12-21
Better range slider for forms: remove empty props in BE

reggie7:
Adjustment for #494
The question here is though: is it ok if we do remove all the missing props on every update? Is it sth desired for all cases? I assume yes, please let me know if that might be false.

### DIFF
--- a/admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
+++ b/admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
@@ -101,7 +101,6 @@ import com.fasterxml.jackson.databind.SerializationFeature;
 import com.peregrine.adaption.PerAsset;
 import com.peregrine.admin.models.Recyclable;
 import com.peregrine.commons.ResourceUtils;
-import com.peregrine.commons.Strings;
 import com.peregrine.commons.util.PerUtil;
 import com.peregrine.rendition.BaseResourceHandler;
 import com.peregrine.replication.ImageMetadataSelector;
@@ -145,7 +144,8 @@ import org.slf4j.LoggerFactory;
 )
 public class AdminResourceHandlerService
     implements AdminResourceHandler {
-    public static final String DELETION_PROPERTY_NAME = "_opDelete";
+    public static final String PN_DELETE_NODE = "_opDelete";
+    public static final String PN_DELETE_PROPS = "_opDeleteProps";
     public static final String MODE_PROPERTY = "mode";
 
     private static final String PARENT_NOT_FOUND = "Could not find %s Parent Resource. Path: '%s', name: '%s'";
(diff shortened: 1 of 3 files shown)


reusr1:
@reggie7 removing all properties that are not sent for save may introduce other issues and would have to be well tested throughout. The lists have a mechanism to remove children that were deleted by keeping an additional list. Maybe that is the better and safer approach here?

reusr1 (changes requested):
@reggie7 @Felix-Puetz it looks like there is an edge case we are not covering in the new slider UI. In https://github.com/headwirecom/themeclean-flex/blob/d79ac8a8053fe0ee202ece93d772ff118f3ff4a0/fragments/teaservertical/model.json#L73-L81 we define a default value for the slider. When we unset the slider the UI will first show the disabled state of the component but after a refresh we show the default value again. Would love to hear your opinions about that how we should handle this case from a UX and front end/back end implementation.

reggie7:
> @reggie7 @Felix-Puetz it looks like there is an edge case we are not covering in the new slider UI. In https://github.com/headwirecom/themeclean-flex/blob/d79ac8a8053fe0ee202ece93d772ff118f3ff4a0/fragments/teaservertical/model.json#L73-L81 we define a default value for the slider. When we unset the slider the UI will first show the disabled state of the component but after a refresh we show the default value again. Would love to hear your opinions about that how we should handle this case from a UX and front end/back end implementation.

@reusr1 I am pretty sure it has to do not with the slider, but with the [`java` model that we generate](https://github.com/headwirecom/themeclean-flex/blob/d79ac8a8053fe0ee202ece93d772ff118f3ff4a0/core/src/main/java/com/themecleanflex/models/TeaserverticalModel.java#L604-L606) based on [`model.json`](https://github.com/headwirecom/themeclean-flex/blob/d79ac8a8053fe0ee202ece93d772ff118f3ff4a0/fragments/teaservertical/model.json#L73-L81):
```
@Inject
@Default(values ="100")
private String textwidth;
```
I'd say the behavior is correct then since the slider simply follows the data that it is fed with. And we should rethink the BE, what do you guys think?

flx-sta:
@reggie7 so you are saying that the java-object grabs the default-value when the value is null?
If that is the case I'd say you are right

reusr1:
I think the problem is in the UX of the slider. @Felix-Puetz if a default value exists and we switch to the off state we should actually move the slider to the default value as well?

flx-sta:
@reusr1 assuming the default value exists on the slider it would be a nice visual feedback cherry

reggie7:
@reusr1 @Felix-Puetz let me guys know if we cover the case below and who (me or @Felix-Puetz) would take care of this.

> I think the problem is in the UX of the slider. @Felix-Puetz if a default value exists and we switch to the off state we should actually move the slider to the default value as well?

Issue #549, opened 2020-09-13
when creating a website as a non admin user we do not check if the website name is available


reusr1:
the current check at https://github.com/headwirecom/peregrine-cms/blob/d9123cdd93d5f317ac70e473127925443c8a8c21/admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/createtenantwizard/template.vue#L199-L213 should call the backend to see if the name is available, we probably should dedupe the call to avoid a lot of requests while typing

reusr1:
@GastonGonzalez we really need to address this one due to kestrel1

GastonGonzalez:
Thanks, @reusr1. I'll take this.

reusr1:
@reggie7 @KWoloszczuk  could you please check if this made it to sling-develop12 and can be closed? thank you

reggie7:
Yes, fixed in #690 and merged. Confirmed now to be sure.

PR #690, opened 2020-12-23
549: check if the site name is available when creating one as a non admin user

reggie7:
Implements #549


admin-base/core/src/main/java/com/peregrine/admin/servlets/AdminPaths.java
```diff
@@ -40,4 +40,5 @@
     public static final String RESOURCE_TYPE_LIST_VERSIONS = API_PREFIX + "listVersions";
     public static final String RESOURCE_TYPE_RESTORE_RECYCLABLE = API_PREFIX + "restoreRecyclable";
     public static final String RESOURCE_TYPE_USER_PREFERENCES = API_PREFIX + "userPreferences";
+    String RESOURCE_TYPE_IS_SITE_NAME_AVAILABLE = API_PREFIX + "sites/name/available";
```
reusr1: any reason not to follow suit and making this constant value `public static final` as the others?

reusr1 (dismissed):
@reggie7 thank you for this PR - we should use Tenant in the backend rather than site. We also should add this endpoint to the swagger api definition

admin-base/core/src/main/java/com/peregrine/admin/servlets/AdminPaths.java
reggie7: Ideally I'd remove the others :) as they are not necessary, but let me simplify the PR and just add the modifiers to the new field for now.

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/createtenantwizard/template.vue
```diff
@@ -99,6 +99,27 @@
...
+                    clearTimeout(timeout)
+                    oldReject()
+                }
+
+                oldReject = reject
+                timeout = setTimeout(later, wait)
+            })
+        }
+    })()
+
```
flx-sta: honestly this whole thing seems out of place. It looks like it's a generalised debouncer instead of a specific one.

So I'd rename it and move it into utils.js (or similar?)

flx-sta (changes requested):
Note: _I didn't look too much into the java part rather then into the js parts_

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/createtenantwizard/template.vue
reggie7: Nice suggestion, done.

flx-sta (approved):

flx-sta (approved):

cmrockwell (commented):
Could not reproduce the issue. Checked-out and deployed the develop-sling12 branch
Signed in as test_user member of all_tenants. My system had a site "test"
When I typed "test" as a new tenant, the error message "name aready in use"
Is this PR still needed?

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/createtenantwizard/template.vue
```diff
@@ -249,7 +270,7 @@
                     }
                 }
 
-                return axios.get('/perapi/admin/sites/name/available?name=' + value)
+                return nameAvailableDebouncer.call(() => axios.get('/perapi/admin/sites/name/available?name=' + value), 375)
```
cmrockwell: Calling `axios.get` in this way looks a bit unusual for this project. It's skipping the API layer. Typically we see functions inside templates calling the api, and then apiImpl calls some variation on fetch to call the API. What we have here is simpler and more direct, but I don't know if it's OK to deviate from the established patterns.

reusr1:
> Could not reproduce the issue. Checked-out and deployed the develop-sling12 branch
> Signed in as test_user member of all_tenants. My system had a site "test"
> When I typed "test" as a new tenant, the error message "name aready in use"
> Is this PR still needed?

I was able to reproduce the issue with the develop-sling12 docker image. log in as admin, create a site ocean and dawn, log in as dawn user, try to create a site called ocean => does not warn

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/createtenantwizard/template.vue
reusr1: agree with @cmrockwell we should go through the api here using axios directly may not include the correct session and may in general cause an issue if/when we switch away from axios.

reusr1:
@cmrockwell I was able to reproduce the issue with the develop-sling12 docker image. log in as admin, create a site ocean and dawn, log in as dawn user, try to create a site called ocean => does not warn

reusr1 (approved):
tested with 2 users and admin, seems to work correctly

Issue #691, opened 2020-12-24
Swagger Editor not available for tenant user


reggie7:
[/perapi/swaggereditor](http://localhost:8080/perapi/swaggereditor/) yields a `404` under a tenant user, but works for admin.

reusr1:
PR opened for this issue, tenant users will now have access to the swagger api (everyone is still declined access)

reusr1:
confirmed issue is resolved in develop-sling12

PR #696, opened 2021-01-04
#691: allow read access to all tenants on /perapi instead of /perapi/admin

reusr1:
I changed the all_tenants/jcr:read from /perapi/admin to /perapi - I don't see any harm in this change (was a jcr:read before as well on sling9 with the previous configs) - would love to hear if everybody agrees or not


reggie7 (approved):

cmrockwell (approved):
Thanks for notifying about this. Based on this, I created a downstream PR and found it included other stuff which I can review for my fork
https://github.com/cmrockwell/peregrine-cms/pull/13/files

PR #699, opened 2021-01-06
FIX #638: rich-toolbar state not updating properly

### LINKED ISSUE #638: rich-toolbar state not updating properly
### Reproduction
* Open Page "article/welcome"
* click on teaser-headline
* click on inline-edit at the bottom (long lorem ipsum text)
* image

### Expected Result
image

flx-sta:
closes #638


reggie7 (approved):

reusr1 (changes requested):
@Felix-Puetz I tried to validate this PR but noticed strange issues when moving from one component to another in inline editing. just by using cursor up/down and moving from one component on the default website home page to another (go from teaser to cards and back) the save dialog is triggered and I can still continue to move between components. Seems there is something seriously wrong here.

flx-sta:
Although it sounds like it has nothing todo with this bug/issue I'll look into it

Found the culprit:
image

flx-sta:
The issue @reusr1 found was actually a major problem.
When switching components the `model` prop from the `component-editor` was not reset but updated.

Meaning when you have 2 components with the following model:
```javascript
// Component 1 - model
{ 
    label: 'test', 
    text: 'lorem ipsum', 
    padding: 10
}
// Component 2 - model
{ 
    label: 'foo', 
    text: 'bar', 
    padding: 5
}
```
It would trigger an update on all 3 keys because the editor-component wasnt re-renderer but updated
So any `$watchers` bound to values of fields would be triggered since a switch from `Component1` to `Component2` would cause e.g. padding to change from `10 -> 5` 

### Additionally
The active element is now reset/blurred when the save-prompt shows up

reggie7 (approved):

flx-sta:
@reusr1 @ehdatheadwire *push*

reusr1 (approved):

reusr1:
saw https://github.com/headwirecom/peregrine-cms/issues/726 while testing this PR but also seems to exist in develop-sling12

PR #708, opened 2021-01-13
Issues/707 - move peregrine build out of docker to prevent travis failures

reusr1:


docker/scripts/install-peregrine.sh
```diff
@@ -2,16 +2,56 @@
...
+  /com.peregrine-cms/external-1.0-SNAPSHOT.zip \
+  /com.peregrine-cms/felib.ui.apps-1.0-SNAPSHOT.zip \
+  /com.peregrine-cms/replication.ui.apps-1.0-SNAPSHOT.zip \
+  /com.peregrine-cms/pagerender-vue.ui.apps-1.0-SNAPSHOT.zip \
+  /com.peregrine-cms/admin.ui.apps-1.0-SNAPSHOT.zip \
+  /com.peregrine-cms/admin.ui.materialize-1.0-SNAPSHOT.zip \
+  /com.peregrine-cms/admin.sling.ui.apps-1.0-SNAPSHOT.zip \
+  /com.peregrine-cms.example/example-vue.ui.apps-1.0-SNAPSHOT.zip \
+  /themeclean/themeclean-ui.apps-1.0-SNAPSHOT.zip \
+  /themeclean-flex/themecleanflex.ui.apps-1.0-SNAPSHOT.zip \
```
cmrockwell: `external-1.0-SNAPSHOT.zip` is gone from develop-sling12

platform/auth/pom.xml
```diff
@@ -162,7 +162,6 @@
         <dependency>
             <groupId>commons-lang</groupId>
             <artifactId>commons-lang</artifactId>
```
cmrockwell: Sling12 doesn't provide `commons-lang` and it's replaced with `commons-lang3`

cmrockwell:
when I run ./builddocker.sh from `develop-sling12` it takes a while but eventually completes successfully.

I checked out this PR
`git fetch upstream pull/708/head && git checkout FETCH_HEAD`
ran ./builddocker.sh

It errors at step 11/14
Screen Shot 2021-02-04 at 1 56 17 PM

cmrockwell (changes requested):
I recommend closing this PR without merging, because after this change docker will only build images using the published packages (i.e. the headwire vagrant develop / Sling9). Docker wouldn't be useful for develop-sling12 PR's going forward.

reusr1:
> I recommend closing this PR without merging, because after this change docker will only build images using the published packages (i.e. the headwire vagrant develop / Sling9). Docker wouldn't be useful for develop-sling12 PR's going forward.

@cmrockwell not sure if that is actually true - it looks like you have to first run the travis build script before doing the docker buildscript - in the process of trying to validate this as well

cmrockwell:
> > I recommend closing this PR without merging, because after this change docker will only build images using the published packages (i.e. the headwire vagrant develop / Sling9). Docker wouldn't be useful for develop-sling12 PR's going forward.
> 
> @cmrockwell not sure if that is actually true - it looks like you have to first run the travis build script before doing the docker buildscript - in the process of trying to validate this as well

Ah ok. In that case the README.md should be updated. See commit 3ab82ed7fbf46bb47d2428664e37677c6bf30dbb

cmrockwell:
I'm not sure whether including the `platform/auth` commons-lang change was a mistake or what, but the build fails. Auth is not included as module of platform, so it needs to be built separately. 

Screen Shot 2021-02-05 at 10 29 23 AM

reusr1:
> > > I recommend closing this PR without merging, because after this change docker will only build images using the published packages (i.e. the headwire vagrant develop / Sling9). Docker wouldn't be useful for develop-sling12 PR's going forward.
> > 
> > 
> > @cmrockwell not sure if that is actually true - it looks like you have to first run the travis build script before doing the docker buildscript - in the process of trying to validate this as well
> 
> Ah ok. In that case the README.md should be updated. See commit [3ab82ed](https://github.com/headwirecom/peregrine-cms/commit/3ab82ed7fbf46bb47d2428664e37677c6bf30dbb)

thank you for updating the readme!

cmrockwell (approved):

reusr1:
I made a lot of changes to this branch after @GastonGonzalez wrote it and @cmrockwell reviewed it. Would be good to open this up for review again, what do you guys think?  

we reduced the travis build time for peregrine from 12 minutes to 8 minutes and the docker image size from 2GB to 800MB

cmrockwell:
> I made a lot of changes to this branch after @GastonGonzalez wrote it and @cmrockwell reviewed it. Would be good to open this up for review again, what do you guys think?
> 
> we reduced the travis build time for peregrine from 12 minutes to 8 minutes and the docker image size from 2GB to 800MB

Looks good! Just added one more thing to the readme  7f506da7d560d0f780fa26effd705e0e119100a4

Issue #658, opened 2020-11-20
Assets do not replicate properly


reggie7:
**Steps to reproduce**
1. Create a new tenant with password and log in as its user,
1. Open Assets of the site,
1. Open browser DevTools,
1. Replicate an image.

**Expected Result**: The asset gets replicated, version is created, icon indicates published status.

**Actual Result**: The asset gets partially replicated, only 1 out of 3 `per:Replication...` properties get set, the globe icon indicates nothing, console displays an error:

![console-log](https://user-images.githubusercontent.com/5563546/99791419-df980b80-2b25-11eb-8681-284a1c93d63f.png)

that comes from [an NPE](https://github.com/headwirecom/peregrine-cms/files/5572968/stacktrace.log):

![NPE](https://user-images.githubusercontent.com/5563546/99791456-f0488180-2b25-11eb-87c0-55d8abf5e664.png)

reggie7:
Found in `develop-sling12`, partially solved in #648 but it still gets deeper than that.

reggie7:
To do: `com.peregrine.slingjunit.localFS.LocalFSJTest#replicateOneAsset`

reggie7:
@KWoloszczuk please test the branch [issues/658-assets-do-not-replicate-properly](https://github.com/headwirecom/peregrine-cms/tree/issues/658-assets-do-not-replicate-properly)

ghost:
Hey @reggie7 I found this while testing your branch:

https://user-images.githubusercontent.com/1230506/107504925-49abd400-6b9c-11eb-89db-a475afde2560.mp4

ehdatheadwire:
@reggie7 is this still in need of PR review? Last note was from 11 days ago so if it's still in need of a review, let us know and we'll push it through so we can close it out.

reggie7:
Yes, PR #773 is prepared and waiting for reviews. It's available for front-facing testing too, so both you @ehdatheadwire and @KWoloszczuk could give it a spin.

ehdatheadwire:
Great, thank you. Will take a look.

PR #710, opened 2021-01-15
Fix: publishing secondary items causes js error

reggie7:
[This commit](https://github.com/headwirecom/peregrine-cms/commit/e005173c26841ffe491fc88f5f69cb34d6c03736) is the actual fix. The rest is cleanup. Let me know if you guys want a PR only for that core part, it will take 5m to prepare :)
The `js` error is gone with this, but there will be an issue present that will be covered by #658 soon, so I'm not fixing it here.


flx-sta (approved):
image
image

cmrockwell (approved):
This one looks ok to me. I wish secondary items could be selected, such that Web Publishing tab could be used for activating and deactivating. But that's a different concern :-)

reggie7:
> 's a different concern :-)

@cmrockwell we can add a ticket for that and evaluate it at some point.

Issue #712, opened 2021-01-20
Move replication code to one place


reggie7:
Replication code lives in 3 places currently:
1. [`admin-base` `core`'s `servlets`](/headwirecom/peregrine-cms/tree/develop/admin-base/core/src/main/java/com/peregrine/admin/servlets),
1. [`platform`'s `core`](/headwirecom/peregrine-cms/tree/develop/platform/base/core/src/main/java/com/peregrine/replication),
1. [`replication` module](/headwirecom/peregrine-cms/tree/develop/platform/replication).

During my work on the topic and various analysis that came with different tickets I came to conclusion that maybe this division is not actually needed. The referencing schema boils down to the following: `platform` → `replication` → `admin`. It's all used inside `admin-base`. `platform` module does not benefit from replication code at all. And actually the split introduces some unnecessary circular dependencies when one wants to use `RepliactionUtil` inside `platform`'s replication code (which prevent one from doing so).
It's all coming back as #658 is in progress. I'd like to cover this change separately though and first move all the replication code to `platform` before coming back to #658.

PR #717, opened 2021-01-21
Remove replication module

reggie7:
Fix for #712 with caveats:
- Code moved to `platform`
- Servlets stay in `admin-base` due to dependencies


cmrockwell:
Hi Reggie. Would it be possible to merge PR719 first (or maybe incorporate it into this one??) I would like remote replication actually working with the UI, so I can test whether anything seems to work properly after this move. Let me know your thoughts. Thanks

reggie7:
@cmrockwell yes, we should do that. And actually I've just merged your branch from #719 into this one here, so the future order of things won't matter.

cmrockwell:
Tried a few times to get this working for the remote use-case.  Junit /system/sling/junit/com.peregrine.slingjunit.author.RemoteReplAuthorJTest.html isn't passing for me.

Comparing to branch issues/718 (remote repl works and junit passes), I see an issue related to the service user.

> 28.01.2021 12:44:35.545 *WARN* [OsgiInstallerImpl] org.apache.sling.jcr.resource.internal.JcrSystemUserValidator The provided service user id 'serviceuser--com.peregrine-cms.base.core--peregrine-distribution-sub-service' is not a known JCR system user id and therefore not allowed in the Sling Service User Mapper.

cmrockwell:
Seem to work with
com.peregrine-cms.base.core:peregrine-distribution-sub-service=distribution-agent-user

image

but the slingjunit test needs that update, and the config below need updating.

http://localhost:8080/bin/browser.html/apps/admin/runmodes/config/org.apache.sling.serviceusermapping.impl.ServiceUserMapperImpl.amended-distributionAgentService

cmrockwell (approved):

platform/base/core/src/main/java/com/peregrine/replication/impl/ReplicationMixinNodeTypesOverride.java
```diff
@@ -75,34 +75,25 @@ private void setup(Configuration configuration) {
...
+        } catch (LoginException e) {
             log.warn("Replication Node Types could not be set", e);
             throw new IllegalArgumentException("Could not obtain Resource Resolver -> setup failed");
         } catch (RepositoryException e) {
             log.warn("Replication Node Types could not be set", e);
             throw new IllegalArgumentException("Could not access Node Types -> setup failed");
-        } finally {
-            if(resourceResolver != null) {
-                try {
-                    resourceResolver.close();
```
cmrockwell, line 104: Why leave the resource resolver open? Long lived resource resolvers are not a good thing.

cmrockwell: Ah nvm it's a try / auto-close thing. Thought that was for input streams. But I guess it works here as well. Have you confirmed it closes?

platform/replication/core/src/main/java/com/peregrine/admin/replication/impl/DistributionReplicationService.java
```diff
@@ -243,8 +242,4 @@ private void setup(Configuration configuration) {
         return answer;
     }
 
-    @Override
-    public List<Resource> prepare(Collection<Resource> resourceList) throws ReplicationException {
-        return new ArrayList<>(resourceList);
-    }
```
cmrockwell: I do not like depending on the default implementation. As we saw, they get changed for whatever reason, and ppl don't check how it affects the other use cases.

platform/base/core/src/main/java/com/peregrine/replication/Replication.java
```diff
@@ -84,8 +85,8 @@ default void prepare(final Resource source, final boolean deep)
     List<Resource> deactivate(Resource source)
         throws ReplicationException;
 
-    default List<Resource> prepare(Collection<Resource> resourceList) throws ReplicationException {
-        return Collections.emptyList();
+    default List<Resource> prepare(final Collection<Resource> resources) throws ReplicationException {
+        return new LinkedList<>(resources);
     }
```
cmrockwell, line 90: I got burned by this. Please get rid of the default implementation. 
List<Resource> prepare(Collection<Resource> resourceList) throws ReplicationException;

cmrockwell:
FYI I added the bug tag, because this PR fixed the bug mentioned in the closed and un-merged PR719 the remote use case

https://github.com/headwirecom/peregrine-cms/pull/719

reggie7:
@cmrockwell we could just as well leave #719 and merge it asap, my guess is that it will grab the necessary approvers much quicker that the PR here.

reusr1 (approved):
as discussed in the watercooler, ok to merge

PR #719, opened 2021-01-21
Implemented prepare method in DistributionReplicationService

cmrockwell:
Here's a fix for the remote replication issue I've been discussing on Slack. Thank you for your consideration


reggie7 (approved):

reggie7:
Nice catch, thanks @cmrockwell

PR #728, opened 2021-02-05
fix #498 logged in as username display issues

### LINKED ISSUE #498: 'logged in as' username display issues
https://github.com/kestrel-1/report-an-issue/issues/1

- [x] when using long user names the display of the user name is not nice
- [x] in the single sign on usecase it would be better to use a profile property such as the email address instead of the actual system user name  

issues/498 branch created

flx-sta:
closes #498


admin-base/materialize/custom/_navbar.scss
```diff
@@ -139,6 +139,53 @@ nav {
...
+            &:first-child {
+              margin-top: 0;
+            }
+
+            &:last-child {
+              margin-bottom: 0;
+            }
+
+            > p {
+              margin: 0 0 0 -5px;
```
reggie7, line 162: Maybe we use `em` or `rem` instead of `px`?

admin-base/materialize/custom/_navbar.scss
```diff
@@ -139,6 +139,53 @@ nav {
...
+              font-weight: bold;
+            }
+          }
+
+          .copy-username {
+            float: right;
+            color: #333;
+            font-size: 13px;
```
reggie7, line 182: Maybe we use `em` or `rem` instead of `px`?

reggie7:
I was thinking - maybe the _copy to clipboard_ icon should be aligned with the **admin** label (the actual thing being copied - username)?

Zrzut ekranu 2021-02-9 o 12 11 00

Apart from that, how often would one use it? Maybe this UI element should show up on hover only?

flx-sta:
> I was thinking - maybe the _copy to clipboard_ icon should be aligned with the **admin** label (the actual thing being copied - username)?
> [image]

not 100% agree with that because I wanted to keep this space and don't take it away for longer names

> Apart from that, how often would one use it? Maybe this UI element should show up on hover only?

That sounds good 👍

admin-base/materialize/custom/_navbar.scss
flx-sta: this is a fix value that should behave the same on all viewports

admin-base/materialize/custom/_navbar.scss
flx-sta: since this is relative to the given font-size: 15px of the text we should keep px here too

flx-sta:
@reggie7 
ezgif com-gif-maker (1)

reggie7 (approved):

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/nav/template.vue
```diff
@@ -133,144 +142,170 @@
...
+      if (breadcrumbs) {
+        return breadcrumbs[0].path.split('/')[4]
+      }
+      return 'welcome'
+    },
+    copyUsername() {
+      this.$refs.usernameInput.setAttribute('type', 'text')
+      this.$refs.usernameInput.select()
+      try {
+        document.execCommand('copy')
```
reusr1: does this really work? I tried to copy the username and then paste it again later but had no success with that

reusr1 (approved):
approving this PR but would like to understand how the copy of username works, also noticed that it apparently copies the email if an email is set up instead of the actual username (according to the toast message)

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/nav/template.vue
flx-sta: @reusr1 ill check again to make sure

flx-sta: There was in-fact a problem with `display: none` I missed. fixed it though and also fixed checking if the copying did work

reusr1 (approved):

flx-sta:
add end-to-end test:https://github.com/peregrine-cms/peregrine-cms-e2e-testing/commit/b20634d5050a9e8853214c7aee1f21b1c20a03df

Issue #578, opened 2020-10-05
Removing component from page does not mark it as updated for replication


reggie7:
Just as the title suggests, deletion of a component within a pge does not trigger replication status icon update
# Steps to reproduce
1. Create a new page,
1. Add an image to the page,
1. Publish all the changes,
1. Remove the image.

**Expected result**: the world icon button indicates that the page got updated and needs re-publishing.
**Actual result**: the world icon button indicates that the page is still published and not updated.

GastonGonzalez:
Good catch, @reggie7.

reggie7:
@reusr1 - reproducible in [`develop`](https://github.com/headwirecom/peregrine-cms/tree/develop) + [`develop`](https://github.com/headwirecom/themeclean-flex/tree/develop).

PR #733, opened 2021-02-08
Add modification time on deleteResource

reggie7:
Fix for #578


admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
```diff
@@ -448,18 +444,20 @@ public DeletionResponse deleteResource(ResourceResolver resourceResolver, String
...
                 throw new ManagementException(String.format(PRIMARY_TYPE_ASKEW_FOR_DELETION, path, primaryType, primaryTypeValue));
             }
 
-            final Resource parent = resource.getParent();
+            final Optional<Resource> parent = Optional.ofNullable(resource.getParent());
             final DeletionResponse response = new DeletionResponse()
                 .setName(resource.getName())
                 .setPath(resource.getPath())
-                .setParentPath(parent != null ? parent.getPath() : "")
+                .setParentPath(parent.map(Resource::getPath).orElse(EMPTY))
```
reusr1, line 458: how is the performance difference of this refactor? I am a bit questioning if this was really necessary and if it improves the performance of the code.

admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
```diff
@@ -448,18 +444,20 @@ public DeletionResponse deleteResource(ResourceResolver resourceResolver, String
...
+
         try {
-            final String primaryTypeValue = resource.getValueMap().get(JCR_PRIMARY_TYPE, EMPTY);
+            final String primaryTypeValue = PerUtil.getPrimaryType(resource);
             if (isNotEmpty(primaryType) && !primaryTypeValue.equals(primaryType)) {
                 throw new ManagementException(String.format(PRIMARY_TYPE_ASKEW_FOR_DELETION, path, primaryType, primaryTypeValue));
             }
 
-            final Resource parent = resource.getParent();
+            final Optional<Resource> parent = Optional.ofNullable(resource.getParent());
```
reusr1, line 454: does this refactor improve the code from a memory footprint perspective?

reusr1 (changes requested):
@reggie7 thank you for this PR - I have a hard time following the fix for the actual issue this PR tries to solve because of all the code refactoring that is going on at the same time. I strongly suggest we stay away from refactoring while fixing an issue due to the low test coverage of the refactored code.

admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
reggie7: Code gets more linear, the number of branches in the flow is reduced below.

admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
reggie7: Code gets more linear, the number of branches in the flow is reduced here.

reggie7:
@reusr1 `parent.ifPresent(baseResourceHandler::updateModification);` is the fix.

admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
reggie7: And we can use a simple line `parent.ifPresent(baseResourceHandler::updateModification);` for the final fix below.

admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
```diff
@@ -436,9 +431,10 @@ private void copyAppsComponentToNewTemplate(String component, Node template) thr
     @Override
     public DeletionResponse deleteResource(ResourceResolver resourceResolver, String path, String primaryType) throws ManagementException {
         final Resource resource = getResource(resourceResolver, path);
-        if (resource == null) {
+        if (isNull(resource)) {
```
flx-sta, line 434: Whats the gain from this?

flx-sta (approved):
Apart from my lack of seeing the gain from using `isNull()` instead of `object == null` I'd say using Optionals at this situation is a good choice.

I remember from the core-devs in my company they did the switch to Optionals too i don't remember all the arguments for it anymore though

**you might wanna wait for a 3rd review**

admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
reggie7: Well, cognitive ease - one can read it like a sentence. I have nothing against algebraic formulas, having math degree, but it's simply easier to read and e.g. not miss the subtle visual difference between `==` and `!=`. It might sound silly, but you know - the less one has to think about irrelevant stuff, the more brain power can be used for the core problem.

Though I'm going to minimize these kind of changes in any code that I write now, people don't want refactoring, got to respect that. So only when it's really needed from now on (not counting existing branches with no PRs yet).

reusr1 (approved):

PR #739, opened 2021-02-12
Simplify aspects of replication FE

reggie7:

We simplify small stuff btw here.
We remove `replicationService` parameter as it's unused and actually incorrect - we should not hardcode `defaultRepl` in FE as it's a name configurable in `OSGi`.
`pending` check is more detailed since at this point it grabbed e.g. `/content/tenant/pages/pending-reviews`.
`replicate` function is flattened a bit.


cmrockwell (approved):
Looks good to me

reusr1 (approved):

PR #742, opened 2021-02-12
Remove redundant stuff in the way of cleaner replication

reggie7:

Preparation cleanups that we base the final solution on. Some redundant stuff is removed, some dead code too.

### DIFF
--- a/platform/base/core/src/main/java/com/peregrine/replication/DefaultReplicationMapperService.java
+++ b/platform/base/core/src/main/java/com/peregrine/replication/DefaultReplicationMapperService.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.function.Function;
 
+import static com.peregrine.commons.Strings.SLASH;
 import static com.peregrine.commons.util.PerUtil.AddAllResourceChecker;
 import static com.peregrine.commons.util.PerUtil.isEmpty;
 import static com.peregrine.commons.util.PerUtil.listMissingResources;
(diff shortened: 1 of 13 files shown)


platform/commons/src/main/java/com/peregrine/commons/TextUtils.java
```diff
@@ -0,0 +1,53 @@
...
+                    throw new IllegalArgumentException(String.format(PLACEHOLDER_UNMATCHED_SEPARATORS, PLACEHOLDER_START_TOKEN, PLACEHOLDER_END_TOKEN));
+                }
+            } else {
+                // Done -> exit
+                break;
+            }
+        }
+        LOGGER.trace("Place Holder handled, return: '{}'", answer);
+        return answer;
+    }
```
cmrockwell, line 51: Good that you moved this out of LocalFileSystemReplicationService. Now we can write regular JUnit tests. Cognitive complexity of this function makes it hard to immediately understand, so Junits could help

cmrockwell: After running debugger on this code, I got a clearer picture about this process. Then I added the test in 208e38e

platform/base/core/src/main/java/com/peregrine/replication/DefaultReplicationMapperService.java
```diff
@@ -131,21 +128,22 @@ void modified(Configuration configuration) {
     private void setup(final Configuration configuration) {
         init(configuration.name(), configuration.description());
         // Register this service as Replication instance
-        logger.trace("Default Mapping: '{}'", configuration.defaultMapping());
-        Map<String, Map<String, String>> temp = splitIntoParameterMap(new String[] {configuration.defaultMapping()}, ":", "\\|", "=");
-        logger.trace("Mapped Default Mapping: '{}'", temp);
+        log.trace("Default Mapping: '{}'", configuration.defaultMapping());
+        var temp = splitIntoParameterMap(configuration.defaultMapping(), ":", "\\|", "=");
```
cmrockwell, line 132: Is it necessary to remove the variables type? It's more informative before. Otherwise difficult to know this is map of maps?

cmrockwell (approved):
After much review. I think this is fine. I hope we can soon CHILL on these refactoring PR's because it doesn't do much for me. It doesn't add test coverage, and takes a long time to review.

reusr1 (approved):

Issue #727, opened 2021-02-05
“No status” after publishing an object


ghost:
Web Publishing tab shows “No status” after publishing an object.

**Steps to Reproduce**

1. Open “objects” Peregrine console.
2. Go to the “tags” folder.
3. Add a tag object.
4. Publish the tag.
5. Open tag properties panel (e.g. with info icon).
6. Open “Web Publishing” tab.

**Expected behavior**
After publishing the status should change to “Published”.

**Actual behavior**
The status is “No Status”.

**Screenshots**

![object no status](https://user-images.githubusercontent.com/1230506/107036514-0e835c80-67ba-11eb-8f60-d3dcd3a56a6e.png)

**Environment**

- `develop-sling12`
- Firefox 85
- Windows 10 (64bit)

ehdatheadwire:
Closing and relocating to enhancements, likely to be included in the overall publishing/unpublishing user flow discussion: https://github.com/peregrine-cms/enhancements/issues/63

reggie7:
#785 fixed it, merged already.

PR #743, opened 2021-02-12
Allow per:Object in ReplicationMixinNodeTypesOverride

reggie7:

Content is fully described in the title.


platform/base/ui.apps/src/main/content/jcr_root/apps/runmodes/config/com.peregrine.replication.impl.ReplicationMixinNodeTypesOverride.xml
```diff
@@ -25,5 +25,5 @@
   -->
 <jcr:root xmlns:jcr="http://www.jcp.org/jcr/1.0" xmlns:sling="http://sling.apache.org/jcr/sling/1.0"
     jcr:primaryType="sling:OsgiConfig"
-    replicationNodeTypes="[per:AssetContent,per:PageContent]"
+    replicationNodeTypes="[per:AssetContent,per:PageContent,per:Object]"
```
reusr1, line 28: are we missing a couple of other types here as well? (nt:file, nt:folder, sling:Folder, per:ObjectDefinition, others?)

cmrockwell:
While I see the notice says success after publishing or unpublishing an object. 

Screen Shot 2021-02-19 at 3 56 01 PM

I also see two problems
1. Web Publishing Tab does not populate publishing information
2. JS errors prevent navigating into nested objects, such as clicking 'allfields' from http://localhost:8080/content/admin/pages/objects.html/path:/content/example/objects 

Is this something to fix here or leave for another PR?

reggie7:
Thanks for seeing into this @cmrockwell, the change is so small that I am sure it's not causing any new issues, just exposing existing ones. We can simply wait with this one for the main PR that I will add next week.

flx-sta (approved):
@reggie7 please open an issue for the follow-up issues

reggie7:
@Felix-Puetz #727 already describes this one:
> 1. Web Publishing Tab does not populate publishing information

@cmrockwell please create an issue for the other case, I don't get the `js` error:
> 2. JS errors prevent navigating into nested objects, such as clicking 'allfields' from http://localhost:8080/content/admin/pages/objects.html/path:/content/example/objects

reusr1 (approved):

platform/base/ui.apps/src/main/content/jcr_root/apps/runmodes/config/com.peregrine.replication.impl.ReplicationMixinNodeTypesOverride.xml
reggie7: Hard to say, depends more on our intentions I guess. I think we could have unwanted side-effects if we add those as well. It requires proper testing and more insight. We could first assess #773's results and view the above configuration from functional perspective.

Issue #568, opened 2020-09-23
Exceptions on develop-sling12 peregrine startup


reggie7:
These are Exceptions that I get when running `java -agentlib:jdwp=transport=dt_socket,address=8001,server=y,suspend=n -jar org.apache.sling.feature.launcher.jar -f com.peregrine-cms.sling.launchpad-12-SNAPSHOT-oak_tar_fds_far.far -D sling.runmodes=author,notshared,oak_tar_fds -p sling` on Windows with the branch [develop-sling12](https://github.com/headwirecom/peregrine-cms/tree/develop-sling12):
```W:\headwire\PeregrineCMS\percms\12>java -agentlib:jdwp=transport=dt_socket,address=8001,server=y,suspend=n -jar org.apache.sling.feature.launcher.jar -f com.peregrine-cms.sling.launchpad-12-SNAPSHOT-oak_tar_fds_far.far -D sling.runmodes=author,notshared,oak_tar_fds -p sling
Listening for transport dt_socket at address: 8001
[INFO]
[INFO] Apache Sling Application Launcher
[INFO] ---------------------------------
[INFO] Initializing...
[INFO] Artifact Repositories: [file:///C:/Users/reggi/.m2/repository, https://repo.maven.apache.org/maven2, https://repository.apache.org/content/groups/snapshots]
[INFO] Assembling provisioning model...
[INFO]
[INFO] Assembling launcher...
[INFO] Using 365 local artifacts, 0 cached artifacts, and 0 downloaded artifacts
[INFO]
[INFO] Starting launcher...
[INFO] Launcher Home: W:\headwire\PeregrineCMS\percms\12\sling
[INFO] Cache Directory: W:\headwire\PeregrineCMS\percms\12\launcher\cache
[INFO]
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.apache.felix.framework.ext.ClassPathExtenderFactory$DefaultClassLoaderExtender (file:/C:/Users/reggi/.m2/repository/org/apache/felix/org.apache.felix.
framework/6.0.3/org.apache.felix.framework-6.0.3.jar) to method java.net.URLClassLoader.addURL(java.net.URL)
WARNING: Please consider reporting this to the maintainers of org.apache.felix.framework.ext.ClassPathExtenderFactory$DefaultClassLoaderExtender
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
[...]

PR #753, opened 2021-02-16
Remove startup exception from ReplicationMixinNodeTypesOverride

reggie7:
Fix for #568

We do not actually need to throw these exceptions.


cmrockwell:
Planning to test this, but do you already know what happens when the mix-in override node types are not defined? Are the replication properties still set on jcr:content?

reggie7:
@cmrockwell just checked - it works indeed when there is no configuration at all. In this case `ReplicationUtil#getReplicationPrimaryNodeTypes` scans node types to check which ones extend `per:Replication` and actually grabs the necessary ones anyway. I'd even say it solves #743 when one removes the configuration.

cmrockwell (approved):

reusr1 (approved):
merged develop-sling12 into this branch again - travis-ci build was stalled

PR #736, opened 2021-02-10
Create paths before adding permissions onto them

reggie7:
Fix for #735

PR #757, opened 2021-02-18
Create paths before adding permissions onto them

reggie7:
Amendment to #736, not sure how we missed that, maybe new paths got added in the meantime.


flx-sta (approved):
nothing spectacular

reggie7:
> nothing spectacular

@Felix-Puetz indeed, that's what happens when you don't refactor 😉

cmrockwell (approved):
Approved. Still catching up. Sorry for the delays

Issue #657, opened 2020-11-20
Tenant user can't replicate themecleanflex assets: NPE is thrown


reggie7:
**Steps to reproduce**
1. Create a new tenant with password and log in as its user,
1. Open Website Themes → ThemeCleanFlex Site → Assets → samples,
1. Open browser DevTools,
1. Replicate e.g. `beach-bird-s-eye-view-daytime-1249586.jpg`.

**Expected Result**: that's actually a good question what the expected result should be. We want the tenant user to be able to replicate referenced assets, but on the other hand it's a shared space.
**Actual Result**: I'm pretty sure though, that the current course of actions is not desired. We get a console error:

![console-error](https://user-images.githubusercontent.com/5563546/99788398-58489900-2b21-11eb-812b-58e35a5439c9.png)

that comes from an [NPE](https://github.com/headwirecom/peregrine-cms/files/5572798/error.log):

![NPE](https://user-images.githubusercontent.com/5563546/99788423-5e3e7a00-2b21-11eb-8635-ca90a325fc9e.png)

reggie7:
Branch: `develop-sling12`

reusr1:
a site has a relation to a theme (in /content/[tenant]). In general, the theme needs to be activated in order for a site to work. I think we should limit replication to a single tenant and not allow cross tenant replication. Best would be to make this an osgi config in the replication settings (cross tenant replication on/off)

reusr1:
@reggie7 before we start addressing this ticket lets come to an understanding of how we are going to solve the issue

I suggest:
- replications are limited to a single tenant and do not cross into other tenants
- we expect that themecleanflex (or the basic theme sites are built from) are replicated when setting up a production environment

This can be of course discussed in a larger group. My main concern: 

if site a uses themecleanflex nd site b uses themecleanflex and they both re published, why would we publish themecleanflex 2x and could this lead to larger issues especially in the future when themeclenflex changes. 

To me the easiest for now is to control the replication of themecleanflex independently from the replication of tenants.

reggie7:
Fixed in [issues/657-tenant-site-setup-replication-NPE](/headwirecom/peregrine-cms/tree/issues/657-tenant-site-setup-replication-NPE). Once [773](https://github.com/headwirecom/peregrine-cms/pull/773#issuecomment-814195222) is merged, I'll open the next PR for this one here.

Issue #720, opened 2021-01-28
Site Setup Replication takes forever in LocalReplicationService


reggie7:
Found in #658. **Site Setup Replication** in `LocalReplicationService` takes too long, I had it work for over 28 minutes. The problem arises from the amount of resources that `replication.findReferences` returns. It's the full tree of the initial list plus its references. There can be even 12k of them. `LocalFileSystemReplicationService` does the same on ~500 resources. I think there's no point in having different implementations of the reference search. Let's fix it.

This is a log from my test, time in seconds (`t` = time since the start, `dt` = time since the previous log):
1. Starting `TenantSetupReplicationServlet.performReplication`, t = 0, dt = 0
2. Starting `extractSiteFeLibs`, t = 0, dt = 0
3. Finished `extractSiteFeLibs`, t = 0, dt = 0
4. Starting `replication.findReferences`, t = 0, dt = 0
5. Finished `replication.findReferences`, t = 1, dt = 1
6. Starting `replication.prepare`, t = 1, dt = 0
7. Finished `replication.prepare`, t = 1, dt = 0
8. Starting `resourceManagement.createVersion`, t = 1, dt = 0
9. Finished `resourceManagement.createVersion`, t = 242, dt = **240**
10. Starting `replication.replicate`, t = 242, dt = 0
11. Finished `replication.replicate`, t = 1720, dt = **1478**
12. Starting `siteMapFilesCache.build`, t = 1720, dt = 0
13. Finished `siteMapFilesCache.build`, t = 1720, dt = 0

reggie7:
@cmrockwell I have noticed similar behavior in **remote replication**, though not so severe, 5 minutes or so.

reggie7:
@reusr1 let's evaluate that after the decision to leave od remove the `local` replication approach.

ehdatheadwire:
@reggie7 and @reusr1 is this required for 1.0.0 release and, if not, should it be moved to the enhancement queue for future work/discussion?

reusr1:
added to 1.0.0 release - 12k resources seems high - do we have dulpicates or items that can not be replicated in the list? Should we limit the site setup replication to the tenant?

reggie7:
@reusr1 it's `local` though, so we need to remember that we were also considering dropping it altogether.

reggie7:
[...]

PR #773, opened 2021-03-01
Fix: assets do not replicate properly (+ side quests found in the process)

reggie7:
Fix for #658 and many more issues with replication found in the implementation process.

All resource types (**page**, **asset**, **object**, **template**) should now allow _publish_ and _unpublish_ operations in `local FS` / `local` / `remote` replication modes (though I would need to check again whether anything was actually broken in the last one). Site Setup Replication should also work for all of them (`local` one takes forever though, I'm not fixing it here).

This issue can both be tested by technical and non-technical reviewers.

In the code review though... It seems like there is a lot of stuff, but there's actually `JUnit` coverage added here, so these are some fresh files in there. Additionally there are many changes in testing tools that were needed to prepare and simplify actual tests. So more than half of the file changes come from testing.
One can easily skip `LocalReplicationService` code review as there are strong plans to simply drop this approach at all (and I haven't written tests for it). @reusr1 would need to confirm that we drop this replication mode.
Other than that I tried to keep the changes to minimum and retract most of the unrelated cleanups, but I didn't revert all small cosmetic ones.

@cmrockwell please confirm that `remote` mode works as expected by yourself.


reggie7:
@cmrockwell thanks to your fix in #831 I could myself already confirm that this PR works as expected within `remote` replication mode.

reggie7:
I've fixed #657 and #720 in [issues/657-tenant-site-setup-replication-NPE](/headwirecom/peregrine-cms/tree/issues/657-tenant-site-setup-replication-NPE) that is based on this PR directly. Once we merge it here, I'll open the next PR.

cmrockwell (approved):
Nice work! I tested using: remote replication ui,  local replication ui,  ran slingjunit remote and local tests,  site setup replication, verified local assets and pages staticreplication, verified version tags are set and removed. Reviewed the code, which looks good to me.
Approved.

flx-sta (approved):
LGTM 🚀

Issue #546, opened 2020-09-10
Unhide child pages from sitemap


GastonGonzalez:
The `excludeFromSitemap` page property currently hides pages set to `true` as well as its children. We need to provide the ability for child pages to be included in the sitemap if desired.

Many sites have the concept of bucket pages that may be used for organizing content. For example:

```
/content
  /pages
    /site
      /posts
        /2020 (bucket) => hide
          /09 (bucket) => hide
            /post1 => allow
            /post2 => allow
            /postN => allow
```

We need the ability to allow child pages under a parent page to be allowed, if desired by the user, to be included in the sitemap. 

There may be a number of ways to solve this problem, so let's discuss the approach here first before starting on the implementation.

Some questions that we should ask ourselves.
1. Should we preserve the existing interface? As it stands, `excludeFromSitemap=true` excludes the current page and all children.
2. This flag is somewhat overloaded as it is also honored by a number components. I believe the pager, but there could be others. We need to think about how the components may be impacted.
3. Should we introduce a second property called `excludePage` that only excludes a single page, but allows the children to be evaluated. In this model, we can still use our current property  `excludeFromSitemap=true` to exclude itself and its children to preserve the behavior used by current Peregrine site. For sites that have bucket pages, they can opt to set `excludeFromSitemap=false` (remove it) and add `excludePage=true` to the bucket pages.

My thought is that `excludeFromSitemap=true` stops the evaluation of child nodes where as `excludePage=true` marks the page to be hidden, but still allows children pages to be evaluated.

reggie7:
[...]

PR #781, opened 2021-03-04
Implement: unhide child pages from sitemap

reggie7:
Implements #546. I am adding a new property `excludeTreeFromSitemap` that works along `excludeFromSitemap`. They are independent now and allow to hide either the page alone or its subtree in site map. It requires content adjustment (I don't see a feasible solution that wouldn't). Requires [180](https://github.com/headwirecom/themeclean-flex/pull/180) to run properly.

### DIFF
--- a/platform/base/core/src/main/java/com/peregrine/sitemap/PageRecognizerBase.java
+++ b/platform/base/core/src/main/java/com/peregrine/sitemap/PageRecognizerBase.java
@@ -36,22 +36,22 @@ public abstract class PageRecognizerBase implements PageRecognizer {
 
     private final String pagePrimaryType;
     private final String pageContentPrimaryType;
-    private final String excludePageFromSiteMapPropertyName;
+    private final String excludeFromSiteMapPropertyName;
     private final String excludeTreeFromSiteMapPropertyName;
 
     protected PageRecognizerBase(
             final String pagePrimaryType,
             final String pageContentPrimaryType,
-            final String excludePageFromSiteMapPropertyName,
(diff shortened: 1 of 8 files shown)


platform/base/core/src/main/java/com/peregrine/sitemap/PageRecognizerBase.java
```diff
@@ -37,17 +37,28 @@
...
 
     public final boolean isPage(final Page candidate) {
+        return isAPageBasically(candidate) && !isExcludedByProperty(excludeFromSiteMapPropertyName, candidate);
+    }
+
+    public final boolean isBucket(final Page candidate) {
+        return isAPageBasically(candidate) && !isExcludedByProperty(excludeTreeFromSiteMapPropertyName, candidate);
+    }
+
+    private boolean isAPageBasically(final Page candidate) {
```
reusr1: I struggle with this method name - not sure what **isAPageBasically** means by reading that name

reusr1 (approved):
approving with the caveat that I really dont like isAPageBasically as a method name

platform/base/core/src/main/java/com/peregrine/sitemap/PageRecognizerBase.java
reggie7: I struggled with finding a good name indeed. I've come up with an alternative `hasAllPageMarkers`, hope will be good enough.

flx-sta (approved):
looks good and the naming issue seems to be resolved. The new one sounds a bit more explanatory

PR #810, opened 2021-03-24
fix for bumping jackson to same version as starter

reusr1:


reusr1:
replaces https://github.com/headwirecom/peregrine-cms/pull/783 as the sling starter is currently 1 version behind

PR #817, opened 2021-03-25
Fix textarea scroll

flx-sta:
I couldn't scroll in textareas on the right side.
this is the fix for it

image


reusr1:
@Felix-Puetz accidentally merged this instead of approving the PR - sorry - but looks good :-)

Issue #660, opened 2020-11-23
Old user gets access to new site


reggie7:
**Steps to reproduce**
1. Grab a fresh instaance with `develop-sling12` on board,
1. Create a new site `Test` with password,
1. Optionally check that you have access as the `test_user`,
1. Remove the site,
1. Create the same site again, this time without any password,
1. Log in as `test_user`.

**Expected Result**: The old `test_user` should not see the new `Test` site / tenant.
**Actual Result**: They do see the new site.

Also - what happens if one does create the site again with the password actually?

reusr1:
@reggie7 I think we should address this issue together with https://github.com/headwirecom/peregrine-cms/issues/447 - we need a way to invite others to collaborate on a site and handle removal of users/sites accordingly

ehdatheadwire:
Judging by your last comment, @reusr1, this might be something to consider as required for 1.0.0 release as we'd like to make it as easy as possible to invite collaborators to the platform and manage users/sites. I'm marking it as such but feel free to adjust or move as warranted.

PR #820, opened 2021-03-26
Issues/660: - make sure old users can not get back into a site with the same name

reusr1:
- [x] admin can not create a site/user pair that already existed in the system
- [x] fix for initial issue in #660

### DIFF
--- a/admin-base/core/src/main/java/com/peregrine/admin/servlets/CreateTenantServlet.java
+++ b/admin-base/core/src/main/java/com/peregrine/admin/servlets/CreateTenantServlet.java
@@ -138,6 +138,7 @@ public class CreateTenantServlet extends AbstractBaseServlet {
         String toTenant = request.getParameter(TO_TENANT_NAME);
         String title = request.getParameter(TO_TENANT_TITLE);
         boolean isAdmin = request.isAdmin();
+
         ResourceResolver resourceResolver = null;
         try {
             logger.trace("Copy Site form: '{}' to: '{}'", fromTenant, toTenant);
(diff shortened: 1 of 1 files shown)


admin-base/core/src/main/java/com/peregrine/admin/servlets/CreateTenantServlet.java
```diff
@@ -158,6 +160,14 @@ protected Response handleRequest(Request request) throws IOException {
...
             String tenantGroupId = toTenant + GROUP_NAME_SUFFIX;
             String tenantUserId = toTenant + USER_NAME_SUFFIX;
+
+            // check if we re trying to create a user that already exists
+            if(isAdmin && userManager.getAuthorizable(tenantUserId) != null) {
+                return new ErrorResponse()
+                .setHttpErrorCode(SC_BAD_REQUEST)
+                .setErrorMessage(FAILED_TO_CREATE_TENANT_SECURITY);
+            }
+
```
reusr1, line 170: we send an error response in case an admin tries to create a site/user pair where the user already exists

reusr1:
if the page can not be created we report a 'permission' issue - since this flow is only for admin users I think the current error is sufficient

reggie7 (approved):

admin-base/core/src/main/java/com/peregrine/admin/servlets/CreateTenantServlet.java
```diff
@@ -158,6 +160,14 @@ protected Response handleRequest(Request request) throws IOException {
             // Create Tenant Group
             String tenantGroupId = toTenant + GROUP_NAME_SUFFIX;
             String tenantUserId = toTenant + USER_NAME_SUFFIX;
+
+            // check if we re trying to create a user that already exists
+            if(isAdmin && userManager.getAuthorizable(tenantUserId) != null) {
+                return new ErrorResponse()
+                .setHttpErrorCode(SC_BAD_REQUEST)
```
flx-sta, line 167: maybe method chain identation missing?

flx-sta (approved):
image
The "workaround" solutions works fine

just 1 question about code-style but thats nothing significant

PR #823, opened 2021-03-27
Disable folders replication in Assets view

### LINKED ISSUE #790: Disable publishing of folders in assets
Related to the listed issue #722. We should disable folder publishing in assets until we work through the desired user flow.

golinski:
Fixes #790.

### DIFF
--- a/admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/explorer/template.vue
+++ b/admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/explorer/template.vue
@@ -361,8 +361,12 @@ export default {
                 $perAdminApp.stateAction('replicate', path)
             },
 
+            isFolder(item) {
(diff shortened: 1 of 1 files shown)


reusr1:
@golinski thank you for the contribution - while reviewing the PR I noticed we have the same issue on all folder replications (I think the other place is in the objects right now). We may want to switch your code just a little bit and instead of limiting the fix to just assets switch it to a more general version like

```
            replicable(item) {
                return !["sling:OrderedFolder", "sling:Folder", "nt:folder"].includes(item.resourceType)
            },
```
what do you think? Let me know if you want to make the change and update the PR or if you want me to change it

golinski:
I can make the change if you wish. I've already noticed that, in fact that was why I decided to add the check if we are in the Assets tab. My reasoning was that for tags we might want to have a "publish all tags in this folder" option.

reusr1 (approved):

reggie7 (approved):

PR #784, opened 2021-03-09
remove old testing framework as well as deprecated sling builder

reusr1:
- we are not using the older testing infrasctrucure anymore
- the sling12 peregrine builder moved to its own project at https://github.com/peregrine-cms/peregrine-builder

PR #747, opened 2021-02-15
Issues/745 remove stale code

reusr1:
this PR removes some unused bundles and brings peregrine up to the latest versions of the corresponding 3rd party bundles.
issues/707 branch is also merged into this PR (new travis build)

please review and merge https://github.com/headwirecom/peregrine-cms/pull/708 (issues/707 PR) before reviewing this PR

PR #831, opened 2021-03-30
Fix remote repl sling12

cmrockwell:
Here's a PR to fix ACL's for remote replication for Sling 12 using Distribution 0.4

References

- https://github.com/headwirecom/peregrine-cms/pull/747/files#diff-1df08ff3e39e133596944dc06cc358c91a06495c1107bd888405a3f19821a379
- https://github.com/headwirecom/peregrine-cms/pull/784

I thought in addition it would be good to update the distribution docs with the current practices.
Understanding that these will be changed soon once percli is updated for Sling12
http://localhost:8080/content/docs/pages/public/distribution.html

Thanks for considering this contribution!

### DIFF
--- a/docs/public/distribution/index.md
+++ b/docs/public/distribution/index.md
@@ -14,10 +14,10 @@ the term Distribution to go along with the Sling parlance.
 
 These are the current supported distributions:
 
-1. Local, in-Peregrine Copies
-1. Remote Sling Distributions (Peregrine to remote Peregrine)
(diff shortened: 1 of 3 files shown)


reusr1 (approved):
@cmrockwell looks great! thank you for the change!

reggie7 (approved):
Awesome, all works fine, thanks a lot @cmrockwell!

PR #834, opened 2021-04-01
fix #832: vfg groups should now behave propery

### LINKED ISSUE #832: Issues with vfg groups in component-panel
> ℹ️ This bug-list is derived from branch TCF branch [`issue/160`](https://github.com/headwirecom/themeclean-flex/tree/issues/160)

### List of known issues:

- [x] Switching any `show <e.g. title>` switch to off and back to on the field pops up in the collapsed `Content` group
    - 
- [x] `Advanced` group -> `semantic element` show an infinte loading animation
    - 

- [x] ~~Using a fresh `teaser-horizontal` component (not example content) you can not open the `content` group~~
    - that was because of all "visible" switches being "off"

flx-sta:
fixed vfg group styling.
- adjustedodd jquery code that caused adding of multiple `display:none` inline-styles.
- only 1 groups is now open (accordeon)

_I tried to add a nice accordeon effect (smooth sliding) but it was laggy/stuttering and didn't work properly due to some elements being hidden with `v-if` so please think about if you really wanted such a feature especially since we are going to switch out vfg soon enough anyway_

closes #832


admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/editor/template.vue
```diff
@@ -209,11 +215,13 @@ export default {
                 const $group = $(this);
                 const $title = $group.find('legend');
                 $title.click(function(e){
-                    $group.find('div').toggle();
-                    $group.toggleClass('active');
+                  const isActive = $group.hasClass('active');
+                  $groups.filter('.active').removeClass('active');
+                  if (!isActive) {
+                    $group.addClass('active');
+                  }
```
flx-sta: this is the jquery code adjustment

admin-base/materialize/custom/_vue-form-generator.scss
```diff
@@ -708,6 +708,34 @@
...
+                    text-align: center;
+                }
+            }
+          }
+
+            &:not(.active) {
+                .form-group {
+                    display: none;
+                }
+            }
```
flx-sta, line 738: this is the visibility adjustment

reusr1:
@Felix-Puetz in general this looks pretty good but it introduces an issue in the image size slider in the cards component:

Screenshot from 2021-04-01 07-39-34

reusr1 (changes requested):
see comment on PR - this seems to introduce an issue in sliders in item lists in a component (see card component)

flx-sta:
I think I've seen this once before already. But couldn't figure out a way to reproduce it and didn't have it appear again.

If you could provide me with a consistent way to reproduce it that would be awesome.

I'll have a look at it on Tuesday 🚀

reusr1:
> I think I've seen this once before already. But couldn't figure out a way to reproduce it and didn't have it appear again.
> 
> If you could provide me with a consistent way to reproduce it that would be awesome.
> 

@Felix-Puetz 

- docker run -it -p 8080:8080 peregrinecms/peregrine-cms:pr-834
- deploy /themeclean-flex[issues/160] to the instance
- create a new website
- click on the cards component on the home page
- try to change the image width

there also seems to be a general issue in collections that the whole collection is draggable (not just the heading of the collection)

flx-sta:
@reusr1 much thanks for the reproduction I was able to spot the issue an resolve it
furthermore some more fixes:

- adjusted the collapse/expand button of the component-explorer. (made 1 dynamic button out of it [code-wise])
- fixed a console error when selecting a filled (e.g. example-variation) card component without having a single card in focus

reusr1:
@Felix-Puetz there seems to be an issue with the groups when going fullscreen with the dialog (fullscreen does not work at all)

flx-sta:
> @Felix-Puetz there seems to be an issue with the groups when going fullscreen with the dialog (fullscreen does not work at all)

resolved those.

Could not fix it "perfectly" though. 
It seems like `$perAdminApp.getView().state.rightPanelFullscreen` reactivity is broken.
So i made a workaround that mirrors the former behaviour of that state but im still updating the original state variable too to not break anything else.
I just don't understand what caused it to be suddenly broken.

reusr1:
@Felix-Puetz I did a quick test and there are some oddities in the fullscreen state and the fullscreen feature does not work anymore after a shift-reload

flx-sta:
@reusr1 Well that was a rapidly fast review.
I'll have another look at it with a fresh mind today 🚀

flx-sta:
@reusr1 I was able to locate the origin of the issue 🚀 
the `setInitialPageEditorState()` method in [apiImpl#1044](https://github.com/headwirecom/peregrine-cms/pull/834/files#diff-e4cf166cd89e4e2c237d6c3da70185c44d21b004c0c5842ccecfa1b04f0c8364R1044) wasn't initialisting the `rightPanelFullscreen` state so the reactivity was broken.
I've added
```javascript
      populateView('/state', 'editorVisible', false)
      populateView('/state', 'rightPanelVisible', true)
 ➕  populateView('/state', 'rightPanelFullscreen', false)
      populateView('/state', 'editor', {})
```
and to be able to set the property inside the workspace vue-component I've used [`computed get/set`](https://github.com/headwirecom/peregrine-cms/pull/834/files#diff-224acea2ca11b3f13930f7546c97afaaac1af2950b703e6df391f7e0feee6bcaR115):
```js
{
  //...
  rightPanelFullscreen: {
      get() {
        return this.state.rightPanelFullscreen;
      },
      set(fullscreen) {
        set($perAdminApp.getView(), '/state/rightPanelFullscreen', fullscreen);
      }
    }
  //...
}
```

reusr1 (approved):
reviewed functionality - seems fine - tried to review the code as much as possible but the change in code style makes this really hard. Suggesting to fix the style before merging the PR

reusr1:
when testing this PR please also use the https://github.com/headwirecom/themeclean-flex/pull/187 version of themeclean flex

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/componentexplorer/template.vue
```diff
@@ -23,161 +23,174 @@
...
-                    const rightName = (right.group + '-' + right.title).toLowerCase();
-                    if(leftName < rightName) return -1;
-                    if(leftName > rightName) return 1;
-                    return 0;
-                })
+export default {
+  props: ['model'],
+  computed: {
+    state() {
+      const state = $perAdminApp.getNodeFromView('/state/componentExplorer');
```
reggie7: I see many semicolons `;` added in a couple of files here. I think we should get rid of them since the prevailing approach of the project is to drop them in `js`.

admin-base/ui.apps/src/main/content/jcr_root/apps/field/collection/template.vue
```diff
@@ -34,15 +34,16 @@
...
+          draggable="false"> {{ item._opDelete }}
         <div
             ref="header"
             class="collapsible-header"
+            draggable="true"
+            v-on:dragstart="onDragStart(item, index, $event)"
+            v-on:dragover.prevent="onDragOver($event, index)"
+            v-on:dragenter.prevent="onDragEnter"
+            v-on:dragleave.prevent="onDragLeave($event, index)"
+            v-on:drop.prevent="onDrop($event, index, item)"
```
reggie7: Don't you prefer `@drop` notation?

reggie7 (approved):

admin-base/ui.apps/src/main/content/jcr_root/apps/field/collection/template.vue
flx-sta: @reggie7 I do indeed but I'd have to refactor all the attributes in the template then

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/componentexplorer/template.vue
flx-sta: sorry my bad. Wrong code-style.
We should in the near future try to have a style XML file that everyone can import into his IDE

flx-sta:
> reviewed functionality - seems fine - tried to review the code as much as possible but the change in code style makes this really hard. Suggesting to fix the style before merging the PR

sorry bout that. Bad habit of hitting the `reformat code`  button

PR #850, opened 2021-04-08
All Replication issues in one spot

reggie7:
#773 + #843 + (#657 + #720) in [issues/657-tenant-site-setup-replication-NPE](/headwirecom/peregrine-cms/tree/issues/657-tenant-site-setup-replication-NPE) as requested by @reusr1


reusr1:
@reggie7 please test this branch a bit - after I tried to run it I noticed the following 2 issues pretty quickly: 

- [x] after doing a full site replication we can not author pages anymore on the author instance
- [ ] the replication state leaks into the publish server (we should not send the replication information to the publisher as we may want to chain publishers in the future)

reggie7:
Closing it as at this point it's basically #843 only.

PR #853, opened 2021-04-10
Make refBy endpoint faster

### LINKED ISSUE #477: performance issue on refby api call
the refby api call ( /perapi/admin/refBy.json ) seems to use a long time once a peregrine environment has a larger number of sites in it. The issue becomes more prominent for users with a large set of allowed websites.

- [x] check if we have an appropriate index for the query
- [x] check if we should limit the call to just the current tenant

golinski:
Fixes #477, although it still a work in progress, maybe a start for a small discussion.

This PR uses an SQL2 (Lucene) query to find nodes that reference the given path. For my sample content tree with ~3000 pages it makes the endpoint almost 10 times faster.

There is a change with respect to the previos solution. For now we do not try to find the name of the property that makes the reference (this needs additional work, as the query itself does not return this info). This begs the question if this is really necessary and the info will be consumed by the caller. The "Pages" view in tenant Admin does not seem to use this additional, very precise, information. It might be a good idea to have two versions of this endpoint one for overviews, the other for more precise info.


reusr1:
@golinski thank you for your PR - I created a new branch in the peregrine-cms repo to make sure your PR builds into a docker image - I did a quick check and looked at the references of the footer template - seems we are getting a bit of a different result than before (we only used to have a reference to the index page by default as that is the only place the footer template is directly used)

Screenshot from 2021-04-12 05-42-01

cmrockwell:
One of the shortcomings of the existing Reference and ReferencedBy is that it missed references within the rich-text component. So if you had built a link to an internal page using the RTE, the traversal method would not count that references. Some time ago, I had a small effort into this problem, so there's a few things you might want to be aware of...

Some of the code added in the PR is duplicate of this.
https://github.com/headwirecom/peregrine-cms/blob/54136a8e061d841b857e234910318038fa6c8a04/platform/base/core/src/main/java/com/peregrine/replication/ReplicationUtil.java#L175-L179

There are a few Sling Junit Tests for getting References using SQL2 queries

- http://localhost:8080/system/sling/junit/com.peregrine.slingjunit.ReferencesJTest.html
- https://github.com/headwirecom/peregrine-cms/blob/develop-sling12/slingjunit.parent/core/src/main/java/com/peregrine/slingjunit/ReferencesJTest.java

golinski:
@reusr1 I've looked ino the issue that too many items are returned. Seems to be connected with the default lucene analyzer used by the SQL2 `contains` constraint. We might need to rollout a special index just for references. I'll look into that.

reggie7:
@golinski I guess we are finding not only `/content/test/templates/footer` valued properties, but also stuff like `/content/test/templates/footer/header-footer`. I think we will need to add something like ` and not contains(*, '%s/')` to the `REFERENCED_BY_QUERY`. Though I've tried it and it does not do the trick as I expected, I need to find out more about `contains` clause.

reggie7:
@golinski can you check [my solution](https://github.com/headwirecom/peregrine-cms/pull/855/files#diff-d8ab419e792db8511e082ae28b47129f469a1047425614820a0aa14978f437b6R90) from the [draft PR](https://github.com/headwirecom/peregrine-cms/pull/855) by @reusr1? It seems to work as expected.

reggie7:
Good job @golinski, let's keep #855 and review your (and mine) work in there. First we need to merge #866 though.

PR #855, opened 2021-04-12
Make refBy endpoint faster

reusr1:
pushed https://github.com/headwirecom/peregrine-cms/pull/853 in to a branch in the repo and created a PR for it


reggie7:
@reusr1 I think I've fixed the [issue](https://github.com/headwirecom/peregrine-cms/pull/853#issuecomment-817779777) that you raised in #853. Could you confirm on your side too?

reggie7:
For some reason GH does not pick up that [`ReferenceListerService`](https://github.com/headwirecom/peregrine-cms/pull/855/files#diff-5285c7053e1d5ee5b84581cf6767631f47674d4ccaf2a99c699a279ba9e32e60) is not a new file, but was moved between packages (it was in `...render...` for some reason). Let's first merge #866 to get a cleaner look here.

golinski:
The default index used by `contains` would put four independent words: `content`, `pages`, `tenant` and `index`. In principle even a completely unrelated string like "Content pages are created by each tenant and an index is created" would match `contains('/content/tenant/pages/index'`), because searched strings are also tokenized into words.

During this week I've tried to get a lucene index working for referencedBy. I was able to create a lucene index inside Oak that collects precisely the necessary data, i.e., puts into the index full paths like `/content/tenant/pages/index`. After exporting the index from Oak, I've confirmed that the index works properly, unfortunately I was unable to call that index from SQL2. The page https://jackrabbit.apache.org/oak/docs/query/lucene.html gives examples of "native queries", but the only example I have found are in the test suite of Jackrabbit Oak and they only concern property indexes (i.e., not accross all properties of a node): https://github.com/apache/jackrabbit-oak/blob/c6ddcc55bee3de915459af01e91edad32d538f3d/oak-search/src/test/java/org/apache/jackrabbit/oak/plugins/index/IndexQueryCommonTest.java#L305.

If we need a precise referencedBy we could leave a `contains` clause only and the filter the returned results in our code. More work to do, in contrast to an existing index, but still less than the original traversal.

reggie7:
@golinski nice catch, let's filter the results of `SQL2` query simply, this way we can also get the referencing property and complete the task to return the same results as before, but much faster.

reusr1:
@reggie7 @golinski since we have new PRs based on this one - should we bring this PR out of draft?

reggie7 (approved):

PR #865, opened 2021-04-19
Issues/852 json schemas for object definitions

flx-sta:
relates to https://github.com/headwirecom/themeclean-flex/pull/181


reusr1:
@Felix-Puetz I was trying to test this out a bit but am having problems when jsonformsvue2vanilla is loaded - seems to be not defined when the form component is initialized (I used the docker image of pr-865 together with the themeclean-flex[jsonforms-for-form-component] branch)

reusr1 (approved):

admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
```diff
@@ -354,6 +355,26 @@ public Resource createObjectDefinition(ResourceResolver resourceResolver, String
         }
     }
 
+    private void addFile(Node parent, String name, String data) throws RepositoryException {
+        Node file = parent.addNode(name, "nt:file");
+        Node fileNode = file.addNode("jcr:content", "nt:resource");
+        fileNode.setProperty("jcr:mimeType", "application/json");
+        fileNode.setProperty("jcr:encoding", "UTF-8");
+        fileNode.setProperty("jcr:data", new StringBufferInputStream(data));
```
reggie7: I would use constants here, but since it's just an extraction, I can let it slip. Though I encourage to use `JcrConstants`, if you want me to - I can replace them for you.

pagerenderer/vue/ui.apps/src/main/content/jcr_root/apps/pagerendervue/structure/page/page.html
```diff
@@ -62,6 +62,8 @@
     </span>
 </div>
 
+<sly data-sly-include="renderer-pre.html"></sly>
```
reggie7, line 65: I don't necessarily enjoy this name. Maybe we call this script differently? E.g. `custom-js` or something similar?

reggie7 (changes requested):

pagerenderer/vue/ui.apps/src/main/content/jcr_root/apps/pagerendervue/structure/page/page.html
reusr1: `renerer-pre.html` indicates that it is executed before `renderer.html` - we could in the future also have a `renderer-post.html` if we choose a different name it would have to support the same idea. I would not enjoy `custom-js` as a name here.

admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
reusr1: agree, this should be changed - @reggie7 would you mind changing this? thank you

pagerenderer/vue/ui.apps/src/main/content/jcr_root/apps/pagerendervue/structure/page/page.html
reggie7: Yeah, it's not so important for me, though I'd ask a native speaker to evaluate it, something simply doesn't sound right for me in this name. I mean - maybe `pre-renderer.html` is more correct? I don't know.

reggie7 (approved):

admin-base/core/src/main/java/com/peregrine/admin/resource/AdminResourceHandlerService.java
reggie7: done

PR #867, opened 2021-04-20
bump sling junit version to 1.1.2

reusr1:
bump sling-junit to 1.1.2 as there are some dependency issues once the launcher is updated to the latest


reusr1:
closing as change was added through a secondary version bump pr

Issue #874, opened 2021-04-20
Site Setup Replication breaks editing in local replication mode


reggie7:
**Steps to Reproduce**
1. Perform **Site Setup Replication** in local replication setup,
2. Edit a page and try to save it.

**Expected Result**
It works.

**Actual Result**
`js` error.

![View](https://user-images.githubusercontent.com/5563546/115443031-05c7f180-a213-11eb-8f3e-c0f1859c29c3.png)

![Console](https://user-images.githubusercontent.com/5563546/115443021-03659780-a213-11eb-80aa-5b2b51b0fb12.png)

PR #878, opened 2021-04-21
Fix update servlet on restricted properties

reggie7:
Fix for #874

### DIFF
--- a/platform/commons/src/main/java/com/peregrine/commons/ResourceUtils.java
+++ b/platform/commons/src/main/java/com/peregrine/commons/ResourceUtils.java
@@ -30,7 +30,6 @@ public final class ResourceUtils {
             JcrConstants.JCR_PREDECESSORS,
             JcrConstants.JCR_UUID,
             JcrConstants.JCR_VERSIONHISTORY,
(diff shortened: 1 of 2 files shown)


platform/commons/src/main/java/com/peregrine/commons/ResourceUtils.java
```diff
@@ -30,6 +30,7 @@
             JcrConstants.JCR_PREDECESSORS,
             JcrConstants.JCR_UUID,
             JcrConstants.JCR_VERSIONHISTORY,
+            JcrConstants.JCR_PREDECESSORS,
```
flx-sta: Isn't this redundant? The same constant is defined above already

flx-sta (changes requested):

platform/commons/src/main/java/com/peregrine/commons/ResourceUtils.java
reggie7: Yes, it is, thank you for noticing :)

flx-sta (approved):
lgtm 🚀  (now)

reusr1 (approved):

PR #892, opened 2021-04-26
Hide un-publish button when resource is not activated

### LINKED ISSUE #890: unpublish an item that was never published throws a client side exception
we should block unpublishing of a never published item or if we allow it at least make sure we dont throw an exception

reggie7:
Fixes #890


flx-sta (approved):

cmrockwell:
Could we use a strike thru? as done for some of the operations

image

To me, seems nicer for the UI to be consistent in how things are disabled. 
WDYT?

flx-sta:
@cmrockwell honestly I do agree with the "consistent design" point but using striketrough is possibly the ugliest way I can think of. rather use color & background-color for making it look disabled.
But I don't really understand what you mean by `"as done for some of the operations"`?

reggie7:
@reusr1 @cmrockwell @Felix-Puetz I have changed the look'n'feel to be consistent throughout the actions and visibly disabled. Let me know what you think.

cmrockwell:
I agree with Felix that strikethrough was not the best way, so thank you @reggie7 ! That looks much nicer imo. 


image

Is it fine that on-click it does nothing when disabled, but when 'Rename' is disabled it has a toast message? I think so.

Does anyone know the hover state BG color? Chrome Inspector was not cooperating this morn. Would like to check the contrast ratio using an online checker. BTW Regular state has [enough contrast](https://webaim.org/resources/contrastchecker/?fcolor=263238&bcolor=ECEFF1)

cmrockwell (approved):

Issue #882, opened 2021-04-23
Block 'unpublish' and 'delete' actions on referenced items


reggie7:

reggie7:
@reusr1 can you confirm my below description, please?
The way I understand it - if an item is referenced somewhere - we don't allow neither to _delete_ it nor to _unpublish_ it?
What about this scenario:
- I add a new page,
- I add a new image,
- I reference the image in the page,
- I publish the image,
- I want to unpublish the image?

I say we would need to allow it then.

golinski:
There are many corner cases here. First, a short scenario

1. An image is added to DAM and is published (because, why not).
2. Similarly an unrelated page is published.
3. On the author instance the image is added to the page, but the updated page is not published yet.
4. We have a problem when trying to unpublish the image: there is a reference, but on the author instance only.

This is not a big deal: author would be notified of the reference and can act accordingly.

The second scenario is worse: we remove a reference on the author that already exists on the publish instance (e.g., we remove an image from a page). Now unpublishing the image would be permitted even though it breaks the live content.

reggie7:
Oh, I know, I just thought of it now - we'll need to use [`VersioningResourceResolver`](https://github.com/headwirecom/peregrine-cms/blob/develop-sling12/platform/base/core/src/main/java/com/peregrine/versions/VersioningResourceResolver.java). It might help us with the proper references for both the above cases. Let's see in practice.

reusr1:
@reggie7 @golinski makes me wonder if we should change the reference search to allow for searching the 'published' version vs the current version

We also may think a bit about not allowing a delete before a node is unpublished or go through the unpublish step if already published.

PR #896, opened 2021-04-28
Block delete action on referenced items

reggie7:
Fixes _delete_ part of #882. Based on #855, so a better changeset view will come once it's merged.
_Unpublish_ part of #882 is a bit trickier, so...


admin-base/core/src/main/java/com/peregrine/admin/servlets/NodesServlet.java
```diff
@@ -232,6 +237,7 @@ private void convertResource(JsonResponse json, ResourceResolver rs, String[] se
                     }
 
                     json.writeAttribute(ANY_DESCENDANT_ACTIVATED, ReplicationUtil.isAnyDescendantReplicated(child));
+                    json.writeAttribute("isReferenced", referenceLister.isReferenced(child));
```
reusr1: not sure if you want to turn this into a constant

reusr1 (approved):
lgtm

admin-base/core/src/main/java/com/peregrine/admin/servlets/NodesServlet.java
reggie7: Done.

flx-sta (approved):

PR #900, opened 2021-04-30
Fix purple version of publish icon

reggie7:
We were missing this:

toolbar

after modification of a resource.


reusr1:
@reggie7 was trying to verify this with the pr-900 docker image but somehow I am not sure what it fixes - I for example changed a published tag and a page property but did not see a difference in the icon - had to reload to get the color change

reggie7:
@reusr1 
1. Publish a page, the earth icon is green
2. Change something on the page - remove / add / change a component, the icon should be purple, it's not currently

flx-sta (approved):

reusr1 (approved):

PR #901, opened 2021-04-30
Fix asset references

reggie7:
We were missing references tab content for assets, I've noticed it while working on #882

### DIFF
--- a/admin-base/ui.apps/src/main/js/stateActions/selectAsset.js
+++ b/admin-base/ui.apps/src/main/js/stateActions/selectAsset.js
@@ -28,17 +28,16 @@ let log = LoggerFactory.logger('selectAsset').setLevelDebug()
 import { set } from '../utils'
 
 export default function(me, target) {
-
     log.fine(target)
-
-    let view = me.getView()
-    return new Promise( (resolve, reject) => {
-        me.getApi().populateComponentDefinitionFromNode('/apps/admin/components/assetview')
+    const api = me.getApi()
+    const view = me.getView()
+    const { selected } = target
+    return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


reusr1 (approved):

admin-base/ui.apps/src/main/js/stateActions/selectAsset.js
```diff
@@ -34,9 +34,11 @@ export default function(me, target) {
...
     return new Promise( (resolve, reject) => {
         me.getApi().populateComponentDefinitionFromNode('/apps/admin/components/assetview')
-        .then( () => {
-            set(view, '/state/tools/asset/show', target.selected)
-            resolve()
+        .then(() => {
+            me.getApi().populateReferencedBy(target.selected).then( () => {
+                set(view, '/state/tools/asset/show', target.selected)
+                resolve()
+            }).catch( error => reject(error))
```
flx-sta: Please take full advantage of the promise chaning by using returns and not nesting `then()` into each other.
Also, probably usefull to return the `target.selected` in the `resolve`. who knows, maybe we use it later
```js
.then(() => {
    return me.getApi().populateReferencedBy(target.selected)
})
.then( () => {
    set(view, '/state/tools/asset/show', target.selected)
    resolve(target.selected)
})
.catch( error => reject(error))
```

flx-sta (approved):
@reggie7 apart from 1 code-style change looks good 🚀

admin-base/ui.apps/src/main/js/stateActions/selectAsset.js
reggie7: Done

PR #911, opened 2021-05-07
Remove cross tenant publishing display in modal window

### LINKED ISSUE #909: Remove cross-tenant publishing display in modal window

reggie7:
Fixes #909, relies on #910 so the latter has to be merged first and this one can be reliably reviewed.

### DIFF
--- a/admin-base/ui.apps/src/main/js/apiImpl.js
+++ b/admin-base/ui.apps/src/main/js/apiImpl.js
@@ -495,8 +495,8 @@ class PerAdminImpl {
         })
   }
 
-  populateReferencedBy(path) {
-    return fetch('/admin/refBy.json' + path)
+  populateReferencedBy(path, sameTenant = false) {
+    return fetch(`/admin/refBy.json${path}${sameTenant ? '?sameTenant=true' : ''}`)
         .then((data) => populateView('/state', 'referencedBy', data))
   }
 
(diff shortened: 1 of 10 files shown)


admin-base/ui.apps/src/main/js/apiImpl.js
```diff
@@ -495,14 +495,14 @@ class PerAdminImpl {
         })
   }
 
-  populateReferencedBy(path) {
-    return fetch('/admin/refBy.json' + path)
+  populateReferencedBy(path, sameTenant = false) {
+    return fetch(`/admin/refBy.json${path}${sameTenant ? '?sameTenant=true' : ''}`)
```
flx-sta: using strings to build querys is a bit unintuitive but the fetch-api is not very advanced on that topic yet.
But you could use this approach: 
```js
fetch('https://example.com?' + new URLSearchParams({
    foo: 'value',
    bar: 2,
}))
```

https://stackoverflow.com/a/58437909

admin-base/ui.apps/src/main/js/apiImpl.js
```diff
@@ -495,14 +495,14 @@ class PerAdminImpl {
...
         .then((data) => populateView('/state', 'referencedBy', data))
   }
 
-  populateReferences(path) {
+  populateReferences(path, sameTenant = false) {
     return new Promise((resolve, reject) => {
-      fetch(`/admin/ref.json${path}`)
+      fetch(`/admin/ref.json${path}${sameTenant ? '?sameTenant=true' : ''}`)
```
flx-sta: same with the params as above

flx-sta (approved):
lgtm except those 2 "code-style improvements"

flx-sta (approved):

reusr1 (approved):

PR #916, opened 2021-05-12
Issues/863 allow file creation in object definitions

### LINKED ISSUE #863: "+ add object definition" doesn't allow creating schema/uischema.json
In the root of level of object definitions the button should stay as it is.

inside a object definiton the button should change its behaviour so that clicking it will end up creating a new file (e.g. `.json`).

# minimal mockup

image

### `onClick`
- opens `editFile` (hopefully vue-coremirror-lite) soon

flx-sta:
## Linked PRs
https://github.com/peregrine-cms/peregrine-cms-e2e-testing/pull/9´

## Screenshots
<kbd></kbd>

<kbd></kbd>

<kbd></kbd>

### DIFF
--- a/admin-base/materialize/custom/_toast.scss
+++ b/admin-base/materialize/custom/_toast.scss
@@ -72,5 +72,14 @@
         content: 'error';
       }
     }
+
+    &.success {
+      background-color: var(--success-bg);
+      color: var(--success-color);
+
+      &:before {
+        content: 'check';
+      }
+    }
   }
 }
--- a/admin-base/materialize/custom/_variables.scss
+++ b/admin-base/materialize/custom/_variables.scss
@@ -10,15 +10,15 @@ $spinner-default-color: var(--pcms-orange);
   --pcms-blue-grey-darken-3: #37474f;
   --pcms-blue-grey: #607d8b;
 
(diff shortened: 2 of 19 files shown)


flx-sta (commented):
I did a code-format on the `constants.js` file but that shouldn't be too big of an issue for th review

reusr1:
@Felix-Puetz at this point we really only want to allow for additional ui-schemas to be created in an object definition. so I think the wizard should represent that and not be a generic file wizard. This wizard is also a departure from all the other creation wizards as it does not have a final confirm step. Since we do expect a json file to be placed we could actually reuse more of the pattern we use when creating a new object and allow for the user to add the json file right during the creation process?

Also feels to me that a name should be sufficient for the entry - don't think we need to force the user to add the json extension, right?

flx-sta:
@reusr1 updated
image

flx-sta:
added tests

reusr1 (approved):
I went ahead and made some minor changes to the labels to better reflect that we are only creating ui schemas in those folders

reusr1:
@Felix-Puetz if you think my changes are acceptable please merge this PR

PR #935, opened 2021-05-26
Issues/862 integrate vue codemirror as editor for object definitions

### LINKED ISSUE #862: integrate vue-codemirror as editor for object-definitions
When clicking `edit` on an object (e.g. json files, js files etc.) it opens the composum editor right now.

We want to change that to vue-codemirror-lite whch is already present as a dependency in the project (`see admin/ui.apps/package.json`)

### LINKED ISSUE #863: "+ add object definition" doesn't allow creating schema/uischema.json

flx-sta:
fixes #862 
fixes #863  

## Linked PRs
#916 
https://github.com/peregrine-cms/peregrine-cms-e2e-testing/pull/10

## Tasks
- [x] enable file-actions
- [x] make bigger files scrollable
- [x] bind `CommandOrControl + S` to `save()` button function
- [x] auto-format / codeformat (`CommandOrControl` + `Alt`  + `l`) 
- [x] missing folder icon
- [x] jumping always to end of document (no matter where I click)
- [x] ~~moving a "page" node is broken~~ (worked perfectly fine for me)
- [x] missing `publish` action
- [x] rename files name regex seems to be wrong
- [x] ~~(i) button in object-definition files doesn't seem to work properly~~ ([worked fine for me](https://github.com/headwirecom/peregrine-cms/pull/935#issuecomment-885666531))


reusr1 (changes requested):
@Felix-Puetz seems there are a couple of issues

- seems you convert the text to json before submitting - I don't think that is necessary as it causes issue when the text in the editor is not a json file
- would be good if the code mirror editor would be the full height
- if I m not mistaken codemirror has code highlighting for json built in

flx-sta:
> @Felix-Puetz seems there are a couple of issues
> 
> * seems you convert the text to json before submitting - I don't think that is necessary as it causes issue when the text in the editor is not a json file

* This now only happens if the answer is content-type application/json

> * would be good if the code mirror editor would be the full height

* Applied this as much as its possible (`setSize('100%', '100%')` didn't work.)

> * if I m not mistaken codemirror has code highlighting for json built in

* Unfortunately this is not that easy in peregrine. There is a `/mode` directory in the codemirror folder where you have to require the necessary modes into your file. I've tried exporting these files into the `/etc/felibs/codemiror/mode/` but I wasn't able to load them async. because neither `import()` nor `require()` seem to be supported. I then tried to import it synchronously by using a static `import` at the top of the file but since you have no access to import files from `/node_modules´ this was unsuccessful either

reusr1:
@Felix-Puetz we need to clean up this page a bit
 - remove the tabs
 - add the breadcrumb above code mirror in the header area (makes it easier to navigate)
 - make sure the right pane shows the regular page info relevant to this file type with the regular actions

we also need a better place for the save button. top right is not what we have anywhere else in peregrine.

flx-sta:
since im implementing a `nt:file` solution for the righ-panel 
- this should also be applied to the secondary files:
  image

- also these files should open in the form-view (codemirror) too

flx-sta:
Current state:
image

flx-sta:
@reusr1 ready for review

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/explorer/template.vue
```diff
@@ -525,11 +528,11 @@ export default {
             },
 
             editable: function(child) {
-                return ['per:Page', 'per:Object'].indexOf(child.resourceType) >= 0
+                return ['per:Page', 'per:Object', 'nt:file'].indexOf(child.resourceType) >= 0
             },
 
             composumEditable: function(child) {
-                return ['nt:file'].indexOf(child.resourceType) >= 0
+                return [].indexOf(child.resourceType) >= 0
```
golinski: `false`?

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/explorerpreviewcontent/template.vue
```diff
@@ -630,12 +647,16 @@ export default {
...
     getSchemaByActiveTab() {
-      if (this.activeTab === Tab.INFO) {
-        return this.getSchema(SchemaKey.MODEL);
-      } else if (this.activeTab === Tab.OG_TAGS) {
-        return this.getSchema(SchemaKey.OG_TAGS);
+      if (this.nodeType === NodeType.FILE) {
+        return this.getGeneratedFileSchema();
       } else {
-        return {};
+        if (this.activeTab === Tab.INFO) {
```
golinski: `else if` on a single line

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/explorerpreviewcontent/template.vue
```diff
@@ -815,10 +841,25 @@ export default {
...
+        
+        if (!to) {
+          const split = this.currentObject.split('/');
+          split.pop();
+          to = split.join('/')
+        }
+
+        $perAdminApp.stateAction('copyFile', {
+          from: this.currentObject, 
+          to
```
golinski, line 853: Is this ok? Looks like it should read `to: to`

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/pageview/template.vue
```diff
@@ -1,60 +1,80 @@
...
   <admin-components-explorerpreviewcontent
-      :key="getBasePath()"
-      :model="model"
-      :tab="tab"
-      :is-edit="isEditPage"
-      :nodeType="NodeType.PAGE"
-      :browserRoot="`${getBasePath()}/pages`"
-      :currentPath="`${getBasePath()}/pages`"
-      :onDelete="onDelete">
+    :model="model"
```
golinski, line 3: `:key` is missing. Is that expected?

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/toolingpage/template.vue
```diff
@@ -113,25 +113,33 @@ export default {
...
-            const node = $perAdminApp.findNodeFromPath($perAdminApp.getView().admin.nodes, target)
-            $perAdminApp.stateAction('editObject', {selected: node.path, path: me.model.dataFrom})
-          } else if (target.startsWith(`/content/${tenant.name}/templates`)) {
-            $perAdminApp.stateAction('editTemplate', target)
+
+          if (resourceType === 'nt:file') {
+            $perAdminApp.stateAction('editFile', {path, resourceType});
           } else {
-            $perAdminApp.stateAction('editPage', target)
+            if (path.startsWith(`/content/${tenant.name}/objects`)) {
```
golinski: `else if` on a single line. Couldn't we use the same logic as the first if (instead of checking `path` look at `resourceType`?

admin-base/ui.apps/src/main/content/jcr_root/content/admin/pages/file/.content.xml
```diff
@@ -0,0 +1,226 @@
...
+      </actions>
+      -->
+      <edit jcr:primaryType="nt:unstructured" sling:resourceType="admin/components/pageview" dataFrom="/state/tools/pages"/>
+
+    </explorer>
+
+    <jsonview jcr:primaryType="nt:unstructured"
+      sling:resourceType="admin/components/debugger" dataFrom="/"/>
+
+    <!-- <tour jcr:primaryType="nt:unstructured"
```
golinski, line 89: Does this commented out fragment serve some purpose?

admin-base/ui.apps/src/main/js/apiImpl.js
```diff
@@ -861,7 +861,7 @@ class PerAdminImpl {
     })
   }
 
-  copyPage(srcPath, targetPath) {
+  copyPage(srcPath, targetPath, name=null) {
```
golinski: Spaces around `=`? I don't know what prettier likes.

admin-base/ui.apps/src/main/js/perAdminApp.js
```diff
@@ -301,97 +301,111 @@ function processLoaders(loaders) {
...
+              } else {
+                logger.error('tenant does not exist or no access to tenant');
+              }
+            });
+          }
+        }
+      }
+    })
+    .then(function() {
+      return api
```
golinski, line 344: Is the value returned from this anonymous function used? Or is this a prettier idea?

admin-base/ui.apps/src/main/js/perAdminApp.js
```diff
@@ -425,17 +439,22 @@ function findActionInTree(component, command, target) {
...
+          'action',
+          command,
+          'not found, ignored, traget was',
+          target
+        );
+      }
     } else {
-        if(component.$parent === component.$root) {
-            if(!findActionInTree(component.$root, command, target)) {
-                logger.error('action', command, 'not found, ignored, traget was', target)
```
golinski, line 433: I know this is probably just reformatted, but:
* target instead of traget
* could we use templated strings?

golinski (approved):

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/explorerpreviewcontent/template.vue
flx-sta: shorthand for `to: to`

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/pageview/template.vue
flx-sta: this was resolved in a different way :)

admin-base/ui.apps/src/main/content/jcr_root/apps/admin/components/toolingpage/template.vue
flx-sta: that would require alot more refactoring. this is basically "legacy support"

admin-base/ui.apps/src/main/content/jcr_root/content/admin/pages/file/.content.xml
flx-sta: idk. its not from me and I don't wanna touch it as I try to focus on the changes rather then cleanups in my ticket

admin-base/ui.apps/src/main/js/perAdminApp.js
flx-sta: typoo fixed. there is advantages of using `...args` syntax here as the object is not transfered to a string in console but rather displayed indepth

flx-sta:
@reusr1 (i) icon works fine in `object-definitions` for me

info-icon-object-definitions

flx-sta:
image

flx-sta (commented):
lgtm

reusr1 (approved):

PR #959, opened 2021-07-13
per:Asset should inherit form nt:file

cmrockwell:
The per:Asset in Peregrine is intended to work with binary files such as images. As such they should inherit from nt:file, and then a per:Asset resource can be adapted to an Inputstream. This is necessary to make use of org.apache.sling.thumbnails

### DIFF
--- a/platform/base/core/src/main/resources/SLING-INF/nodetypes/asset.cnd
+++ b/platform/base/core/src/main/resources/SLING-INF/nodetypes/asset.cnd
@@ -25,7 +25,7 @@
 <'per'='http://www.peregrine-cms.com/jcr/cms/1.0'>
 <'sling'='http://sling.apache.org/jcr/sling/1.0'>
 
-[per:Asset] > nt:hierarchyNode, mix:created, mix:lastModified, sling:Resource
+[per:Asset] > nt:hierarchyNode, mix:created, mix:lastModified, sling:Resource, nt:file
   primaryitem jcr:content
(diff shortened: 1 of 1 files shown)


reusr1:
@cmrockwell not sure if we really should make per:Asset depend on nt:file - an asset does not need to be a file, right? its more the sub node that holds the content that would need to be a file

cmrockwell:
`nt:file` isn't literally a file. it's a JCR node meant to represent storage of a file. 

> Nodes of this node type may be used to represent files. This node type inherits the child nodes and properties of nt:hierarchyNode and requires a single child node called jcr:content. The jcr:content node is used to hold the actual content of the file.

It's the same thing as per:Asset as for as I can tell. When per:Asset inherits from nt:file, then Sling Thumbnail can process per:Assets, but currently the following method returns null. 

```
@Override
public InputStream getThumbnail(Resource resource) {
    return resource.adaptTo(InputStream.class);
}
```

Only `nt:file` and `nt:resource` types can be adapted to an InputStream. So Sling Thumbnails could only process per:Assets if we approve this PR.

In addition, we might also make per:AssetContent inherit from nt:resource because this has same relationship to nt:file as per:AssetContent does to per:Asset. 

https://docs.adobe.com/content/docs/en/spec/jcr/1.0/6.7.22.6_nt_file.html
https://docs.adobe.com/content/docs/en/spec/jcr/1.0/6.7.22.9_nt_resource.html

PR #1088, opened 2022-07-19
Moved the Distribution Repo-Init to its package

schaefa:
With Sling-12 our original setup fails
This PR moves the setup for the distribution from the Peregrine to the Distribution repo-init


schaefa:
I encountered a similar issue with the per:FeLib and moved the creation of this node type to the Felib repo-init and with it the creation of the /etc/felibs folder.
In addition the update of the sling:target on the root is now done with repo-init as well to make sure it will be update if the property is already set.

platform/base/ui.apps/src/main/content/jcr_root/apps/runmodes/config/org.apache.sling.jcr.repoinit.RepositoryInitializer-peregrine.config
```diff
@@ -20,19 +20,23 @@ scripts=[\
...
 
 # Create the necessary folders to apply permissions below otherwise the Launchpad will fail during startup
 
     create path (sling:Folder) /apps
     create path (sling:Folder) /content
     create path (sling:Folder) /etc
-    create path (sling:Folder) /etc/felibs
+    create path (sling:Folder) /var
     create path (sling:Folder) /etc/packages
+    create path (sling:Folder) /var
```
golinski, line 36: Dow we need this twice (see line 34)?

golinski (approved):

platform/base/ui.apps/src/main/content/jcr_root/apps/runmodes/config/org.apache.sling.jcr.repoinit.RepositoryInitializer-peregrine.config
schaefa: Fixed

PR #1094, opened 2022-07-29
1011 - Added Object Model Support for JSon Schemas

schaefa:
This is a PR to add support for Object Definitions based on JSon Schemas
This PR supersedes the PR for 852


schaefa:
Done

PR #1097, opened 2022-08-05
1096 - Enable Actions on Objects

schaefa:
Also fixed an isuse with copying a resource where the name/title is ‘null’ instead of null


schaefa:
Done

